Issue #2, opened 2015-10-30
http://demo.erlang.org/eep.html is not accessible ?


crquan:
```console
$ curl -v http://demo.erlang.org/eep.html
* Could not resolve host: demo.erlang.org
* Closing connection 0
curl: (6) Could not resolve host: demo.erlang.org
```

dmitriid:
Should be `http://erlang.org/eep.html`

MrMormon:
Would someone with the proper permissions please fix the repo description to correct this?

RaimoNiskanen:
Done by merging Pull req #3 and adding one commit

PR #3, opened 2016-01-18
Correct URL

brucify:
demo.erlang.org is not available anymore.

See https://github.com/erlang/eep/issues/2


RaimoNiskanen:
Not complete; there are two more references in EEP 0.  I'll fix them in a new commit

PR #6, opened 2018-09-10
Update EEP 49 (eep-0049.md)

ferd:
Add content on comparisons with Elixir's `with`, and expand on not
supporting all ok- and error-based tuples following community feedback


RaimoNiskanen:
You introduced a HTTP link that I will linkify om master right now...

PR #16, opened 2020-11-30
Update EEP-54 according to the decision of OTB

bjorng:
The OTP Technical Board approved the EEP, but requested a few
changes:

* The optional `module` and `function` keys in the map will point
  the module and function for a `format_error` function.

* The arguments for `format_error` has been simplified to just
  two arguments: The exception reason and the first entry in
  stacktrace.

Taken together, those to changes makes it easy to have one module
with a `format_error/2` that can explain errors for multiple
modules.

### DIFF
--- a/eeps/eep-0054.md
+++ b/eeps/eep-0054.md
@@ -3,7 +3,7 @@
     Type: Standards Track
     Created: 14-Sep-2020
     Erlang-Version: 24
-    Post-History:
+    Post-History: 14-Oct-2020, 27-Nov-2020
 ****
 EEP 54: Provide more information about errors
 ----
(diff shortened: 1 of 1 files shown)


essen:
Why pass only the first entry to the stacktrace and not the full stacktrace? The format_error function can extract the first entry on its own.

Saying this because the first entry in the stacktrace may not necessarily be the function the user was calling, it may be a second internal function, and a user format_error function may want to look at the second entry instead.

bjorng:
> Why pass only the first entry to the stacktrace and not the full stacktrace? The format_error function can extract the first entry on its own.

After an internal discussion in the OTP team, we have decided to adopt your suggestion.

PR #21, opened 2021-03-17
Update EEP 56

Maria-12648430:
This PR updates EEP 56 with a link to the reference implementation and a new paragraph about forward compatibility.

### DIFF
--- a/eeps/eep-0056.md
+++ b/eeps/eep-0056.md
@@ -4,7 +4,7 @@
     Type: Standards Track
     Created: 04-Mar-2021
     Erlang-Version: 25
-    Post-History: 08-Mar-2021, 17-Mar-2021
+    Post-History: 08-Mar-2021, 17-Mar-2021, 23-Mar-2021
     Replaces:
 ****
 EEP 56: Automatic supervisor shutdown triggered by termination of significant children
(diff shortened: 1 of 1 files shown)


eeps/eep-0056.md
```diff
@@ -158,11 +158,31 @@ the new child spec and supervisor flags are optional and default to values
 that result in the current behavior. Also, all the current workarounds
 outlined in the Motivation will still work.
 
+A concern exists however in regards to _forward_ compatibility. As the changes
+proposed in this document do not change or remove any existing but only
+introduce new keys for supervisor flags and child specs, it is possible to
+compile and run applications that utilize the significant child behavior even
+with OTP versions that predate this document and thus do not implement it.
+Such an application compiled with older OTP versions will leak processes,
```
zuiderkwast, line 166: I'm not sure "forward compatibility" is the right term for this. This is a *backward* compatibility problem IMHO, but not for supervisor itself but for the applications using the new feature. If an application uses this new feature on an older OTP version, it will "almost work" i.e. it will leak processes, so it's not fully compatible with the older OPT versions. The supervisor extension itself is backward compatible though (since all programs written for older OTP versions will work in newer OTP versions with this extension).

Forward compatibility, if my imagination is correct, refers to the concern that this version will not be compatible with future versions or the attempt to facilitate for future versions to extend/modify it in a backward compatible way.

Maria-12648430: Thanks for the feedback :) But what do you think the proper term for this kind of issue is? ^^; I have no idea...

zuiderkwast: I don't know. Just describe it? e.g. "Although this enhancement is backward compatible, applications using this enhancement may not be compatible with previous OTP versions unless proper care is taken...." and the explanation you already wrote.

Maria-12648430: I updated the passage according to your suggestions (with some small changes). Better now? :)

RaimoNiskanen:
Thank you for your Pull Request!

PR #23, opened 2021-04-26
eep54: Add general and reason to return map

garazdawi:
The `general` return key was added in order to allow information
to be given about the default I/O device in [io:format("hello")][io_format_1].
It also allows third-party error_report implementations (such as Elixir)
much more freedom in what they can return.

The `reason` return key was added in order for third-party error_report
implementations (such as Elixir) to influence what is printed to describe
the actual error.

[io_format_1]: http://erlang.org/doc/man/io.html#format-1

### DIFF
--- a/eeps/eep-0054.md
+++ b/eeps/eep-0054.md
@@ -118,6 +118,13 @@ that was in error, there should be a map element with the argument number
 as the key (that is, `1` for the first argument, `2` for the second, and so on)
 and a `unicode:chardata()` term as the value.
 
+The atoms `general` and `reason` can also be returned in the map.
+`general` represents a generic error that is not attributed for a specific argument
+(for example badarg of io:format("Hello") when the default device is dead).
+`reason` will tell the error pretty printer to print the returned string instead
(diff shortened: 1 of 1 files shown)


eeps/eep-0054.md
```diff
@@ -118,6 +118,12 @@ that was in error, there should be a map element with the argument number
 as the key (that is, `1` for the first argument, `2` for the second, and so on)
 and a `unicode:chardata()` term as the value.
 
+The atoms `general` and `reason` can also be returned in the map.
+`general` represents a generic error that is not attributed for a specific argument
+(for example badarg of io:format("Hello") when the default device is dead).
+`reason` will tell the error pretty printer to print the returned string instead
```
zuiderkwast, line 124: Are the values of `reason` and `general` of type `unicode:chardata()` like the other values in the map? It could be clarified.

zuiderkwast (commented):
Nice!

eeps/eep-0054.md
garazdawi: done

PR #24, opened 2021-06-02
New EEP: Module attributes for documentation

josevalim:


eeps/eep-0057.md
```diff
@@ -0,0 +1,235 @@
...
+
+    -doc #{
+      authors => [<<"José Valim"/utf8>>]
+    }.
+
+For this reason, the Build and Packaging Working Group of the Erlang Ecosystem Foundation has also discussed the option of making Unicode binary strings easier to write in Erlang, such as `u"Héllò World"`. While adding such construct is not a direct part of this EEP, the point made in this subsection is that, if we require the documentation or its metadata to be a binary, we may need to provide conveniences for writing UTF-8 binary strings in Erlang.
+
+## Triple-quoted strings
+
+In the examples above, we have used regular Erlang strings for the documentation. However, they come with a limitation: escaping can become confusing quite fast as soon as the documentation itself contains double quoted strings. One option here is to add triple-quoted strings to Erlang:
```
DianaOlympos, line 164: I think the triple quote part here is a bit redundant and not obvious. Maybe just put something saying this is an example and keep bringing the triple quote only after this example ?

eeps/eep-0057.md
```diff
@@ -0,0 +1,235 @@
...
+
+    -doc "Remove double-quotes (\") from the given string".
+    remove_double_quotes(String) ->
+      ...
+
+For this reason, we may opt-in to provide triple-quoted strings, which reads better in multi-line format and reduces the need for escaping:
```
josevalim, line 170:
```suggestion
Another feature often related to structure documentation is triple-quoted strings. To understand why they may be useful, consider the following example that has a double quote in its documentation:

    -doc "Remove double-quotes (\") from the given string".
    remove_double_quotes(String) ->
      ...

For this reason, we may support triple-quoted strings, which reads better in multi-line format and reduces the need for escaping:
```

eeps/eep-0057.md
```diff
@@ -0,0 +1,235 @@
...
+
+Rationale
+=========
+
+The main limitation in EDoc today is that the documentation is kept as code comments. This requires an explicit tool to parse said code comments, which complicates access to the docs by IDEs, from the shell, etc. There have been recent improvements in this area by making EDoc compile to EEP 48 but it still requires an explicit step.
+
+Furthermore, the “code comments” approach is more complex implementation wise, as it requires parsing the source code alongside code comments, parsing the code comments, and so on. Personally, I also enjoy the explicit distinction between documentation and code comments: they have different requirements and different audiences.
+
+This EEP proposes the addition of two module attributes to Erlang: `-doc` and `-moduledoc`. This EEP also includes an extra section on additional features that can help writing documentation with those attributes more productive, but that's optional. As with EEP 48, this proposal pertains exclusively to API references and their documentation. It doesn't cover guides, tutorials, and other documentation formats.
```
josevalim, line 22:
```suggestion
This EEP proposes the addition of two module attributes to Erlang: `-doc` and `-moduledoc`. This EEP also includes an extra section on additional features that can make writing documentation more productive, but they are optional and some of them should likely be explored more in depth in other EEPs. Still, including them here can be relevant to provide a long term view and some of challenges found when providing structured documentation.

As with EEP 48, this proposal pertains exclusively to API references and their documentation. It doesn't cover guides, tutorials, and other documentation formats.
```

eeps/eep-0057.md
```diff
@@ -0,0 +1,235 @@
...
+    -doc "
+    Decodes the given binary from base64.
+    ".
+    -spec decode(binary()) -> {ok, binary()} | error.
+    encode(Binary) ->
+      % ....
+
+The new `-moduledoc` attribute can be listed anywhere and it will contain the documentation for the given module. The `-doc` attribute must be listed once before each function and it will contain the documentation for the following function.
+
+Listing multiple `-doc` attributes for the same function should raise an exception.
```
josevalim, line 50:
```suggestion
Listing multiple `-doc` attributes for the same function should warn or error accordingly.
```

eeps/eep-0057.md
```diff
@@ -0,0 +1,235 @@
...
+    -moduledoc "
+    Convenience functions for encoding and decoding from base64.
+    ".
+    -moduledoc #{cross_references => [binary]}.
+
+A list of built-in attributes is available on EEP 48.
+
+## Compilation
+
+Compiling a module with the `-moduledoc` or `-doc` attributes will automatically generate a Docs chunk into its .beam file, making the documentation directly accessible in the shell. A `-nodocs` flag can be added to `erlc` to skip the generation of the chunk.
```
josevalim, line 122:
```suggestion
Compiling a module with the `-moduledoc` or `-doc` attributes will automatically generate a Docs chunk into its .beam file, making the documentation directly accessible in the shell. A `-nodocs` flag can be added to `erlc` to skip the generation of the chunk.

Release tools should also prune the Documentation chunk out of .beam files by default. Note this is already done by `beam_lib:strip_release/1` and `beam_lib:strip_files/1`.
```

eeps/eep-0057.md
```diff
@@ -0,0 +1,235 @@
...
+
+EDoc and source code documentation
+==================================
+
+## Should documentation be in the source code?
+
+There is one discussion which is the debate if API Reference documentation should exist in the source code or be placed in a separate file. There are arguments against and in favor of both approaches. If the documentation is alongside the source code, it is easier to keep it up to date. At the same time, it may pollute the source code. This proposal does not advocate that one is better than the other. Rather, because we already have a tool that colocates documentation and source code, namely EDoc, this EEP aims to propose a more integrated variant of said tool.
```
josevalim, line 209:
```suggestion
There is one discussion which is the debate if API Reference documentation should exist in the source code or be placed in a separate file. There are arguments against and in favor of both approaches. If the documentation is alongside the source code, it is easier to keep it up to date. At the same time, it may pollute the source code. This proposal does not advocate that one is better than the other. Rather, because we already have a tool that colocates documentation and source code, namely EDoc, this EEP aims to propose a more integrated variant of said tool, which will hopefully lead developers to better document their Erlang modules and libraries.
```

essen:
First, attributes without parens is weird. :-)

I agree about wanting a different solution than code comments. Comments are not documentation. I disagree about documentation in the source code being easier to keep in sync, it gets out of date just as easily as external documentation. On the other hand when the documentation is in the source code it's easier to sneek in potentially breaking documentation changes unnoticed.

As a result I would much prefer to keep external documentation but still make use of these attributes, perhaps via something like:

```
-moduledoc({external, "doc/src/manual/my_module.asciidoc"}).
-doc({external, "doc/src/manual/my_module.my_function.asciidoc"}).
```

This would close any debate about documentation inside or outside the source code since both would be possible. There is also no issues with regarding to quotes when using external documentation.

I suppose a compile flag could be added to locate the external files more easily, though `.` is perfectly fine in my case.

eeps/eep-0057.md
```diff
@@ -0,0 +1,237 @@
...
+Similarly, if the goal is to eventually phase out EDoc, the Erlang/OTP team needs to discuss what happens with the existing EDoc to XML conversion in `erl_docgen`. Two options exist:
+
+  * Generate EDoc to XML one last time and use the XML as the source from now on
+
+  * Convert EDoc to the documentation style in this proposal and teach erl_docgen how to convert the Docs chunk to XML
+
+Copyright
+=========
+
+This document has been placed in the public domain.
```
hauleth, line 228: This probably should strictly describe license, like CC0, as not all jurisdictions allow voluntary assigning document to public domain.

josevalim: This is the standard of all other EEPs, so I won't change it, but perhaps this conversation needs to brought up elsewhere? Or maybe it only matters for what is valid in Sweden in this case?

josevalim:
@essen thank you for the comments and the suggestion. I will amend the proposal to allow embedding external documents too. It may be a bit repetitive to provide a path for each doc but I guess someone could write a parse transform that maps everything in a directory to moduledoc+doc if they really want to reduce the duplication?

eeps/eep-0057.md
hauleth: @josevalim created #25

KennethL:
another attribute 'include_doc' could be introduced for inclusion of doc
written in other files than the source code file. That would make it
possible to mix doc written in the source file with doc written outside in
a flexible way.

eeps/eep-0057.md
richcarl: AFAIK, the templates were copied from PEP, so this phrasing is likely just an old artefact. I agree it should be changed to CC0. In many countries (including most of Europe) saying "this is placed in the public domain" does not have any legal effect. Copyright is automatic as of the Berne convention of 1988.

RaimoNiskanen:
An EEP should have line breaks as any Markdown document (that follows E-Mail conventions).
From EEP 33:
> You must adhere to the Emacs convention of adding two spaces at the
> end of every sentence.  You should fill your paragraphs to column 70,
> but under no circumstances should your lines extend past column 79.
> If your code samples spill over column 79, you should rewrite them.

eeps/eep-0057.md
```diff
@@ -0,0 +1,349 @@
...
+parsing the code comments, and so on.  Personally, I also enjoy the
+explicit distinction between documentation and code comments: they have
+different requirements and different audiences.
+
+This EEP proposes the addition of two module attributes to Erlang: `-doc`
+and `-moduledoc`.  This EEP also includes an extra section on additional
+features that can help writing documentation with those attributes more
+productive, but that's optional. As with EEP 48, this proposal pertains
+exclusively to API references and their documentation.  It doesn't cover
+guides, tutorials, and other documentation formats.
```
josevalim, line 39:
```suggestion
This EEP proposes the addition of two module attributes to Erlang: `-doc`
and `-moduledoc`.  It also includes an extra section on additional features
that can make aid writing documentation, but they are optional and some
of them should likely be explored more in depth in other EEPs.
Still, they are included to provide a long term view of the features and
challenges related to structured documentation.
          
As with EEP 48, this proposal pertains exclusively to API references and
their documentation.  It doesn't cover guides, tutorials, and other
documentation formats.
```

josevalim:
@RaimoNiskanen updated!

@essen and @KennethL, I added a section about supporting `-doc({file, Path})`. There is only one concern: you wouldn't be able to place documentation metadata in an asciidoc file. But I assume that is fine? The metadata is typically License+Author for modules and Since+Deprecated annotations for functions.

lhoguin:
Yes no problem about metadata.

eeps/eep-0057.md
```diff
@@ -0,0 +1,353 @@
...
+    Decodes the given binary from base64.
+    ".
+    -spec decode(binary()) -> {ok, binary()} | error.
+    encode(Binary) ->
+      % ....
+
+The new `-moduledoc` attribute can be listed anywhere and it will contain
+the documentation for the given module.  The `-doc` attribute must be
+listed once before each function and it will contain the documentation
+for the following function.
```
ferd, line 73: Any ordering preferences compared to type specs?

eeps/eep-0057.md
```diff
@@ -0,0 +1,353 @@
...
+      authors => [<<"José Valim"/utf8>>]
+    }.
+
+For this reason, the Build and Packaging Working Group of the
+Erlang Ecosystem Foundation has also discussed the option of making
+Unicode binary strings easier to write in Erlang, such as
+`u"Héllò World"`.  While adding such construct is not a direct part
+of this EEP, the point made in this subsection is that, if we require
+the documentation or its metadata to be a binary, we may need to
+provide conveniences for writing UTF-8 binary strings in Erlang.
```
ferd, line 252: This, to me, would point towards supporting both formats and converting. Erlang strings should represent safe unicode data that can be converted later with `unicode:characters_to_binary(CharData)` and for the time being provides safe unicode with low syntactic overhead.

eeps/eep-0057.md
josevalim: @ferd please expand? I am not sure I follow the question. :)

eeps/eep-0057.md
josevalim: Yeah, I think I agree with you. What about metadata? Should it be automatically rewritten too?

eeps/eep-0057.md
ferd: People have a tendency to do something like:

```
%% @doc This function turns the programmer into a much more productive one
-spec levelup(programmer()) -> programmer().
levelup(Programmer) -> Programmer*10.
```

The new literal translation would be:

```
-doc "This function turns the programmer into a much more productive one".
-spec levelup(programmer()) -> programmer().
levelup(Programmer) -> Programmer*10.
```

This will work with the spec above if we read "listed _once before_ each function", meaning there can be one `-doc` attribute before any function and it will get assigned to the one that follows.

However the spec could also force the following if we read it to mean "the doc must be listed once _right before_ each function":

```
-spec levelup(programmer()) -> programmer().
-doc "This function turns the programmer into a much more productive one".
levelup(Programmer) -> Programmer*10.
```

eeps/eep-0057.md
ferd: It could make sense. I'd see this as an extension of all Erlang-style things where `CharData` gets flattened down for you when possible. It would in principle not harm anything as long as you specify you expect data to be UTF-8 compatible, meaning raw strings, binaries in the utf8 encoding, or lists of either of these.

eeps/eep-0057.md
josevalim: Ah, I see. We don't look at other attributes, so it can be anywhere before a function. I will clarify.

hauleth:
@josevalim

>  you wouldn't be able to place documentation metadata in an asciidoc file

ASCIDoc has built-in support for document metadata via [attributes](https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#attributes-and-substitutions). So with "smart enough" implementation the metadata could be stored in the external files as well.

essen:
I would suggest keeping the external file format implementation not smart to begin with and build upon them later if there's strong interest. Asciidoc attributes are one option, but there's also Hugo-like header variables that work with any file format for example.

moiseev:
Any thoughts on the warning similar to `-compile(warn_missing_spec).` now that these docs are known to the compiler?

eeps/eep-0057.md
josevalim, line 73:
```suggestion
The new `-moduledoc` attribute can be listed anywhere and it will contain
the documentation for the given module.  The `-doc` attribute must be
listed anywhere before a function and it will contain the documentation
for the following function.  For instance, the example below:

    -doc "Example".
    -spec example() -> ok.
    example() -> ok.

is equivalent to:

    -spec example() -> ok.
    -doc "Example".
    example() -> ok.    
```

eeps/eep-0057.md
```diff
@@ -0,0 +1,353 @@
...
+    encode(Binary) ->
+      % ....
+
+The new `-moduledoc` attribute can be listed anywhere and it will contain
+the documentation for the given module.  The `-doc` attribute must be
+listed once before each function and it will contain the documentation
+for the following function.
+
+Listing multiple `-doc` attributes for the same function should warn or
+error accordingly.
```
josevalim, line 76:
```suggestion
Listing multiple `-doc` attributes for the same function should warn or
error accordingly, unless the documentation is being set to hidden.
For example, this is valid:

    -doc "Example".
    -doc hidden.
    example() -> ok.

But this should warn/error:

    -doc "Example".
    -doc "Updated example".
    example() -> ok.

This as well:

    -doc "Example".
    example(one) -> 1;
    -doc "Updated example".
    example(two) -> 2;
```

eeps/eep-0057.md
```diff
@@ -0,0 +1,353 @@
...
+The doctest attribute will access the documentation entries in
+the base64 Docs chunk, extract all of the examples, and run them.
+Of course, there is nothing stopping doctests from being implemented
+on top of EDoc today, but the fact the documentation is already
+parsed by the parser/compiler, makes doctests trivial to add.
+Doctests would benefit from a separate EEP, as there are some extra
+considerations, as doctesting exceptions, unparseable formats, etc,
+but it is worth mentioning them given their benefits to users and
+documentation authors.
+
```
josevalim, line 313:
```suggestion

## `warn_missing_doc`

Another optional feature is to support a `-compile(warn_missing_doc).`
that emits a warning whenever an exported function is not documented.

```

eeps/eep-0057.md
josevalim, line 252:
```suggestion
For this reason, the Build and Packaging Working Group of the
Erlang Ecosystem Foundation has also discussed the option of making
Unicode binary strings easier to write in Erlang, such as
`u"Héllò World"`.  While adding such construct is not a direct part
of this EEP, the point made in this subsection is that, if we require
the documentation or its metadata to be a binary, we may need to
provide conveniences for writing UTF-8 binary strings in Erlang.
Alternatively, this may also indicate the best option is to support
the documentation to be written as charlists (or `unicode:chardata()`)
and convert them to UTF-8 encoded binaries internally.
```

josevalim:
@moiseev I have added a section mentioning this feature could also be added in the future too.

eeps/eep-0057.md
```diff
@@ -0,0 +1,386 @@
...
+is equivalent to:
+
+    -spec example() -> ok.
+    -doc "Example".
+    example() -> ok.    
+
+Listing multiple `-doc` attributes for the same function should warn or
+error accordingly, unless the documentation is being set to hidden.
+For example, this is valid:
```
hauleth, line 87: This probably should be:

```suggestion
Listing multiple `-doc` attributes with string for the same function
should warn orerror accordingly, unless the documentation is being
set to hidden.
For example, this is valid:
```

To not conflict with the metadata section.

ferd: `orerror` sounds like a typo though

hauleth: @ferd fixed.

hauleth:
> I would suggest keeping the external file format implementation not smart to begin with and build upon them later if there's strong interest.

@essen that was what I meant. It shouldn't be Erlang concern to understand and extract metadata from documentation, but the documentation processor (external) should be able to achieve something like that in future.

michalmuskala:
This seems like a fantastic proposal that would simplify dealing with in-code documentation significantly compared to edoc comments. Is there anything I could do to help move this forward?

KennethL:
Can you please change the file name to eep-0059.md and I will merge this PR so we can refer to it as EEP-59

josevalim:
@KennethL done! I have also updated some parts with the feedback received in this PR.

josevalim:
I pushed again to make the linter happy!

PR #28, opened 2021-06-17
New EEP: Delayed restarts of supervisor children

Maria-12648430:
We use `57` as temporary EEP number here, aware that there are already two other candidates for it. The number will be changed accordingly if and when this EEP has been accepted.

### DIFF
--- a/eeps/eep-0057.md
+++ b/eeps/eep-0057.md
@@ -1,0 +1,227 @@
+    Author: Maria Scott <maria-12648430(at)hnc-agency(dot)org>
+            Jan Uhlig <juhlig(at)hnc-agency(dot)org>
+    Status: Draft
+    Type: Standards Track
+    Created: 16-Jun-2021
+    Erlang-Version: 25.0
+    Post-History:
+    Replaces:
+****
+EEP 57: Delayed restarts of supervisor children
+----
+
+
+
+Abstract
+========
+
+This EEP introduces a mechanism to delay the restart of crashed children of
+supervisors.
+
+
+
+Motivation
+==========
+
+At the time of this writing, crashed children of supervisors will be
+restarted immediately. If they fail to restart, another restart
+attempt is made, again immediately, until the either the restarting
+succeeds or the supervisors restart intensity limit is exhausted.
+
+This approach is rather aggressive, and there are scenarios where
+frantic restarting is unfeasible and will likely achieve nothing but
+a shutdown of the supervisor.
+For example, the child in question may depend on an external service
+like a database which in turn may be temporarily down or overloaded.
+
+The request for a way to introduce restart delays has appeared on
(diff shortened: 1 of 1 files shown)


eeps/eep-0057.md
```diff
@@ -0,0 +1,227 @@
...
+rest_for_one, we can even say that the delays between retries are either the
+same or decreasing. The children before the offending child are running and so
+need not be considered. The children behind the offending child have not been
+tried, and having used the maximum delay in (2), they need not be considered also.
+Put simply, only the delay of the offending child needs to be considered when
+retrying a restart.
+
+A special case in rest_for_one is that one of the still running children may
+crash while children further back are in delay. In this case, the calculation of
+the new delay should take the remainder of the already running delay into consideration.
```
josevalim, line 155: Suggestion: make it explicit that the delayed children and the newly failed children will be grouped.

```suggestion
crash while children further back are in delay. In this case, the new failed children
will be grouped with the currently delayed children, and the calculation of the new
delay should take the remainder of the already running delay into consideration.
```

Maria-12648430: Thanks :)

Maria-12648430:
Following [the discussion on the eeps mailing list](http://erlang.org/pipermail/eeps/2021-June/000727.html) in which (mainly) @ferd voiced valid concerns, we decide to retire this EEP.

lhoguin:
Weak.

I understand the concerns, and perhaps this is not the appropriate solution, but I don't believe "roll your own via gen_statem" is appropriate either. We are lacking a component in OTP that would sit between the supervisor and the process that must be delayed.

Perhaps now that we are able to change a `gen_statem` module (via `change_callback_module`), there could be a `gen_statem` added to OTP that would deal with delays, retries, backoff, jitter and whatever other common strategy for dealing with these issues. When the process starts, it goes through your own `gen_statem` as normal. When the connection (or other resource) is eventually dropped, you switch to the `wait_gen_statem` module (for lack of better name). It could have different options to select the strategy. Then once the wait is done, `wait_gen_statem` switches back to your own module.

This approach would mean that the state data must be shared between the two. A convention could determine how `wait_gen_statem` can access its own state I suppose.

This approach would only work for `gen_statem` processes, but that's probably fine as converting a `gen_server` to a `gen_statem` is trivial, and processes that need to do this are most likely already implemented using `gen_statem`.

The `wait_gen_statem` module would either have to reply to calls with an error, or to defer them for when the wait is done. Perhaps it can be made configurable.

What do you think?

Maria-12648430:
> What do you think?

I'll think about it :)

zuiderkwast:
I think the EEP should be merged even if it is (or will be) rejected, for the record. It's a proposal that has been sent out to the mailing list already.

PR #32, opened 2021-09-14
Map comprehensions EEP

seriyps:
Follow-up from https://github.com/erlang/otp/pull/4856

### DIFF
--- a/eeps/eep-0058.md
+++ b/eeps/eep-0058.md
@@ -1,0 +1,255 @@
+    Author: Sergey Prokhorov <seriy(dot)pr(at)gmail(dot)com>
+    Status: Draft
+    Type: Standards Track
+    Created: 14-Sep-2021
+    Erlang-Version: 25
+    Post-History:
+****
+EEP 58: Map comprehensions
+----
+
+
+Abstract
+========
+
+This EEP proposes syntax and semantics for map comprehensions and
+generators, similar to list and binary comprehensions and generators,
+but operating on maps and map iterators.
+
+Copyright
+=========
+
+This document has been placed in the public domain.
+
+
+Specification
+=============
+
+Maps and initial map comprehension syntax and semantics were proposed
+in EEP-43. List and binary comprehensions were specified post-fact
+in EEP-12. This EEP is based on those two EEPs.
+
+Syntax construct commonly referred as "comprehension" is in fact
+consists of 3 semi-independent parts:
+
+- "generator" - is the expression that iterates over the container,
+  matches each element of the container over pattern and binding
+  variables which can be used in "comprehensions" and "filters".
+- "comprehension" - is the expression that is executed for each
+  element produced by generator that successfully matched the pattern
+  and passed the "filter" check. Value produced by this expression
+  is appended to the resulting generated container.
+- "filter" - is the expression that is evaluated for each element
+  produced by generator that matched the pattern and returns a
+  boolean result: when it returns `false` the current element is
+  skipped from being passed to "comprehension"; when it results
+  in `true`, the element is passed to "comprehension".
+
+So, in this document we are going to keep this separation, but
+sometimes calling the whole combination of comprehension, filter and
+generator as just "comprehension", because generator syntax is not
+used outside of the comprehension context.
+
+
+### Syntax
+
+Proposed syntax for map comprehension part is
+
+    '#{' KeyExpr '=>' ValueExpr '||' GeneratorsFilters '}'
+
+Where `GeneratorsFilters` are one or more list/map/binary generator
+and/or filter, separated by comma.
+
+Proposed syntax for map generator part is
+
+    KeyPattern ':=' ValuePattern '<-' MapOrIteratorExpr
+
+Example combination of map comprehension and map generator
+
+    #{Key => Value || Key := Value <- MapOrIterator}
+
+Map comprehension may be combined with list or binary generator and
+vice-versa
+
+    [{Key, Value} || Key := Value <- MapOrIterator]
+    #{K => K || K <- List}
+    #{K => V || <<K, V>> <= Binary}
+
+`MapOrIterator` is any Erlang expression that evaluates to `map()`
+or `maps:iterator()` datatype.
+
+### Semantics
+
+Map comprehension should be semantically equivalent to the
+`maps:from_list/1` call over the list comprehension producing
+the list of 2-tuples:
+
+    #{K => V || ...}
+
+equivalent to
+
+    maps:from_list([{K, V} || ...])
+
+
+Map generator should be semantically equivalent to the list generator
+running over the result of `maps:to_list/1`, when the input is a
+map datatype:
+
+    #{ ... || K := V <- Map}
+
+equivalent to
+
+    #{ ... || {K, V} <- maps:to_list(Map)}
+
+And for `maps:iterator()` as input - it can't be easily expressed as
+equivalent list generator. But the idea is that it's equivalent to
+the consumption of the iterator by repetitive calls to `maps:next/1`
+and matching `Key` and `Value` over key and value patterns
+accordingly on each iteration loop.
+
+Map generator should raise `error({bad_generator, ExprResult})`
+if the `MapOrIterator` expression evaluation produced anything but
+map or map iterator.
+
+Key pattern is allowed to be catch-all / contain unbound variables
+(which is not allowed in other map pattern-matching contexts).
+
+Variable binding and shadowing rules should be the same as for list
+and binary comprehensions (which follows from the equivalence rules
+described above).
+
+Rationale
+=========
+
+### Syntax for arrow, generator pattern and comprehension key-value pair
+
+We decided to go with the syntax proposed in EEP-43.
+
+`#{ .. }` were chosen for the outer tokens of map comprehension
+because the same are used when constructing a new map - rule similar to
+list and binary comprehensions.
+
+`K := V <-` is not ambiguous for the parser, so unlike binary
+generator, there is no need for a new arrow. `<-` were chosen over
+`<=` because latter is less visually distinguishable when it is
+placed relatively close to the `=>` and `:=` tokens:
+
+    #{K => V || K := V <= Map}.
+    % vs
+    #{K => V || K := V <- Map}.
+
+`KeyPattern := ValuePattern` is the same syntax that is used in
+map pattern-matching (however, in map generators KeyPattern is also
+allowed to be a match-all / unbound variable).
+
+`KeyExpr => ValueExpr` is the same syntax that is used for the new
+map construction.
+
+### Order in which map generator yields key-value pairs
+
+Since maps do not have any order specified, the order in which
+generator would yield key-value pairs should not be guaranteed (same
+way as the ordering of the list, resulting from `maps:to_list/1` is
+not guaranteed).
+
+### Behavior when comprehension produced duplicate keys
+
+Same as `maps:from_list/1`: the latter (right-most) value is used
+and the previous values are ignored. It might be less predictable
+when map comprehension is used in combination with map generator
+(because map generator order is not defined), but is important when
+map comprehension is used with list of binary generator.
+
+### Whether to allow map iterator as input for map generator
+
+Map iterators are allowed in `maps:filter/2`, `maps:fold/3` and
+`maps:map/2`. Since comprehensions are often used in the same
+contexts as those functions, it makes sense to allow map iterators
+so comprehensions could be used as a drop-in replacement for the
+aforementioned functions.
+
+### Map updates via map comprehension
+
+Should map updates be allowed with map-comprehensions? That is should
+this be valid syntax:
+
+    M#{ K => V*2 || K := V <- M}
+
+or this
+
+    M#{ K := V*2 || K := V <- M}
+
+There is no final decision about how useful might it be.
+
+More-or-less the same result may be achieved using `maps:merge/2`
+function (with less room for optimization).
+
+One (probably minor) argument against is the fact that "map update"
+syntax is a common reason of very confusing errors programmers make
+when they forget to add a comma between two maps:
+
+    > [#{a => b}
+       #{a => d}].
+    #{a => d}
+
+Here "map update" syntax have been accidentally created instead of
+intended "list of two maps".
+
+There exists some risk that similar issue might happen with
+"update via comprehension":
+
+    my_fun(Map) ->
+        Options = #{a => b}
+        #{ K => V || K := V <- smth(Map, Options) }.
+
+However probability is lower, since mistake usually happens in literal
+definitions, but programmers rarely use comprehensions in literal
+definitions.
+
+Motivation
+==========
+
+Having comprehensions and generators for lists and binaries, it
+seems logical that there should be a generator for maps as well.
+For example, Python language has versions of comprehensions for
+both lists and dicts (but also generators in comprehensions work for
+any iterable).
+Another factor is that it's a quite common practice in the code
+seen in a wild to use combinations of `maps:from_list/1`,
+`maps:to_list/1` with list comprehensions and list high-order
+functions in a situations when high-order functions from `maps`
+module are not flexible enough (eg, we want to convert map not to
+another map, but to some different structure, or to modify both
+key and value). But use of `maps:from/to_list` might be a problem
+for the performance since it is eager and creates garbage on heap.
+
+Backwards Compatibility
+=======================
+
+All the existing Erlang code won't be affected, since the syntax
+extensions we are proposing in this EEP are not valid syntax in
+Erlang right now.
+Erlang code written with map comprehensions would be non-parseable
+by the older Erlang compiler.
+Some of the tools working with intermediate Erlang forms (eg, AST)
+have to be updated (eg, parse-transforms). However, we expect
+core-erlang form to not be affected - similar to list and binary
+comprehensions.
+
+Reference Implementation
+========================
+
+The reference implementation is provided in a form of pull-request
+on GitHub
+
+    https://github.com/erlang/otp/pull/4856
+
+
+[EmacsVar]: <> "Local Variables:"
+[EmacsVar]: <> "mode: markdown"
+[EmacsVar]: <> "indent-tabs-mode: nil"
+[EmacsVar]: <> "sentence-end-double-space: t"
+[EmacsVar]: <> "fill-column: 70"
+[EmacsVar]: <> "coding: utf-8"
+[EmacsVar]: <> "End:"
+[VimVar]: <> " vim: set fileencoding=utf-8 expandtab shiftwidth=4 softtabstop=4: "
--- a/eeps/eep-0058.md
+++ b/eeps/eep-0058.md
@@ -25,42 +25,42 @@ This document has been placed in the public domain.
 Specification
 =============
 
-Maps and initial map comprehension syntax and semantics were proposed
+Maps and the initial map comprehension syntax and semantics were proposed
 in EEP-43. List and binary comprehensions were specified post-fact
 in EEP-12. This EEP is based on those two EEPs.
 
-Syntax construct commonly referred as "comprehension" is in fact
-consists of 3 semi-independent parts:
+The syntax construct commonly referred as a "comprehension" is, in fact,
+composed of 3 semi-independent parts:
 
-- "generator" - is the expression that iterates over the container,
-  matches each element of the container over pattern and binding
-  variables which can be used in "comprehensions" and "filters".
-- "comprehension" - is the expression that is executed for each
-  element produced by generator that successfully matched the pattern
-  and passed the "filter" check. Value produced by this expression
+- the "generator" - is the expression that iterates over the container,
+  matches each element of the container over a pattern and binds the
+  variables which can be used in the "comprehensions" and the "filters".
+- the "comprehension" - is the expression that is executed for each
+  element produced by the generator that successfully matched the pattern
+  and passed the "filter" check. The value produced by this expression
   is appended to the resulting generated container.
-- "filter" - is the expression that is evaluated for each element
-  produced by generator that matched the pattern and returns a
-  boolean result: when it returns `false` the current element is
-  skipped from being passed to "comprehension"; when it results
-  in `true`, the element is passed to "comprehension".
+- the "filter" - is the expression that is evaluated for each element
+  produced by the generator that matched the pattern and returns a
+  boolean result: when it returns `false`, the current element is
+  skipped from being passed to the "comprehension"; when it results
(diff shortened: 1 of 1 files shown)


eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+Copyright
+=========
+
+This document has been placed in the public domain.
+
+
+Specification
+=============
+
+Maps and initial map comprehension syntax and semantics were proposed
```
pablocostass, line 28:
```suggestion
Maps and the initial map comprehension syntax and semantics were proposed
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+in EEP-43. List and binary comprehensions were specified post-fact
+in EEP-12. This EEP is based on those two EEPs.
+
+Syntax construct commonly referred as "comprehension" is in fact
+consists of 3 semi-independent parts:
```
pablocostass, line 33:
```suggestion
The syntax construct commonly referred as a "comprehension" is, in fact,
composed of 3 semi-independent parts:
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+- "generator" - is the expression that iterates over the container,
+  matches each element of the container over pattern and binding
+  variables which can be used in "comprehensions" and "filters".
```
pablocostass, line 37:
```suggestion
- the "generator" - is the expression that iterates over the container,
  matches each element of the container over a pattern and binds the
  variables which can be used in the "comprehensions" and the "filters".
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+- "comprehension" - is the expression that is executed for each
+  element produced by generator that successfully matched the pattern
+  and passed the "filter" check. Value produced by this expression
+  is appended to the resulting generated container.
```
pablocostass, line 41:
```suggestion
- the "comprehension" - is the expression that is executed for each
  element produced by the generator that successfully matched the pattern
  and passed the "filter" check. The value produced by this expression
  is appended to the resulting generated container.
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+- "filter" - is the expression that is evaluated for each element
+  produced by generator that matched the pattern and returns a
+  boolean result: when it returns `false` the current element is
+  skipped from being passed to "comprehension"; when it results
+  in `true`, the element is passed to "comprehension".
```
pablocostass, line 46:
```suggestion
- the "filter" - is the expression that is evaluated for each element
  produced by the generator that matched the pattern and returns a
  boolean result: when it returns `false`, the current element is
  skipped from being passed to the "comprehension"; when it results
  in `true`, the element is passed to the "comprehension".
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+So, in this document we are going to keep this separation, but
+sometimes calling the whole combination of comprehension, filter and
+generator as just "comprehension", because generator syntax is not
+used outside of the comprehension context.
```
pablocostass, line 51:
```suggestion
So, in this document we are going to keep this separation, but
sometimes will refer to the whole combination of the comprehension, the filter and
the generator as a "comprehension", because the generator syntax is not
used outside of the comprehension context.
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+
+### Syntax
+
+Proposed syntax for map comprehension part is
```
pablocostass, line 56:
```suggestion
The proposed syntax for a map comprehension is
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+    '#{' KeyExpr '=>' ValueExpr '||' GeneratorsFilters '}'
+
+Where `GeneratorsFilters` are one or more list/map/binary generator
+and/or filter, separated by comma.
```
pablocostass, line 61:
```suggestion
Where `GeneratorsFilters` are one or more list/map/binary generators
and/or filters, separated by a comma.
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+Proposed syntax for map generator part is
```
pablocostass, line 63:
```suggestion
The proposed syntax for the map generator part is
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+Example combination of map comprehension and map generator
+
+    #{Key => Value || Key := Value <- MapOrIterator}
+
+Map comprehension may be combined with list or binary generator and
+vice-versa
+
+    [{Key, Value} || Key := Value <- MapOrIterator]
+    #{K => K || K <- List}
+    #{K => V || <<K, V>> <= Binary}
```
pablocostass, line 76: I still think this is super cool :)

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+`MapOrIterator` is any Erlang expression that evaluates to `map()`
+or `maps:iterator()` datatype.
+
+### Semantics
+
+Map comprehension should be semantically equivalent to the
+`maps:from_list/1` call over the list comprehension producing
+the list of 2-tuples:
```
pablocostass, line 85:
```suggestion
A map comprehension should be semantically equivalent to the
`maps:from_list/1` call over a list comprehension producing
a list of 2-tuples:
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+    #{K => V || ...}
+
+equivalent to
```
pablocostass, line 89:
```suggestion
is equivalent to
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+    maps:from_list([{K, V} || ...])
+
+
+Map generator should be semantically equivalent to the list generator
+running over the result of `maps:to_list/1`, when the input is a
+map datatype:
+
+    #{ ... || K := V <- Map}
+
+equivalent to
```
pablocostass, line 100:
```suggestion
is equivalent to
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+We decided to go with the syntax proposed in EEP-43.
+
+`#{ .. }` were chosen for the outer tokens of map comprehension
+because the same are used when constructing a new map - rule similar to
+list and binary comprehensions.
+
+`K := V <-` is not ambiguous for the parser, so unlike binary
+generator, there is no need for a new arrow. `<-` were chosen over
+`<=` because latter is less visually distinguishable when it is
```
pablocostass, line 134:
```suggestion
generator, there is no need for a new arrow. `<-` was chosen over
`<=` because the latter is less visually distinguishable when it is
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+map pattern-matching (however, in map generators KeyPattern is also
+allowed to be a match-all / unbound variable).
+
+`KeyExpr => ValueExpr` is the same syntax that is used for the new
+map construction.
+
+### Order in which map generator yields key-value pairs
+
+Since maps do not have any order specified, the order in which
+generator would yield key-value pairs should not be guaranteed (same
```
pablocostass, line 151:
```suggestion
Since maps do not have any order specified, the order in which the
generator would yield key-value pairs should not be guaranteed (same
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+way as the ordering of the list, resulting from `maps:to_list/1` is
+not guaranteed).
+
+### Behavior when comprehension produced duplicate keys
+
+Same as `maps:from_list/1`: the latter (right-most) value is used
+and the previous values are ignored. It might be less predictable
+when map comprehension is used in combination with map generator
+(because map generator order is not defined), but is important when
+map comprehension is used with list of binary generator.
```
pablocostass, line 161:
```suggestion
Same as `maps:from_list/1`: the latest (right-most) value is used
and the previous values are ignored. It might be less predictable
when a map comprehension is used in combination with a map generator
(because the map generator order is not defined), but is important when
a map comprehension is used with a list or a binary generator.
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+    M#{ K := V*2 || K := V <- M}
+
+There is no final decision about how useful might it be.
+
+More-or-less the same result may be achieved using `maps:merge/2`
+function (with less room for optimization).
+
+One (probably minor) argument against is the fact that "map update"
+syntax is a common reason of very confusing errors programmers make
```
pablocostass, line 188:
```suggestion
One (probably minor) argument against it is the fact that the "map update"
syntax is a common reason for very confusing errors programmers make
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+when they forget to add a comma between two maps:
+
+    > [#{a => b}
+       #{a => d}].
+    #{a => d}
+
+Here "map update" syntax have been accidentally created instead of
+intended "list of two maps".
```
pablocostass, line 196:
```suggestion
Here the "map update" syntax has been accidentally created instead of
the intended "list of two maps".
```

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+Motivation
+==========
+
+Having comprehensions and generators for lists and binaries, it
+seems logical that there should be a generator for maps as well.
+For example, Python language has versions of comprehensions for
+both lists and dicts (but also generators in comprehensions work for
+any iterable).
+Another factor is that it's a quite common practice in the code
+seen in a wild to use combinations of `maps:from_list/1`,
```
pablocostass, line 218:
```suggestion
seen in the wild to use combinations of `maps:from_list/1`,
```

essen:
@pablocostass Could you please use the review button next time so we get all your comments as a single email instead of twenty? Just a suggestion, no need to reply. And sorry for the additional noise everyone.

zuiderkwast (commented):
Nice! I'm looking forward to this missing piece of the language.

Just some minor comments. Also note that the EEP guideline says something about multiple blank lines before every heading. You may want to look it up.

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+More-or-less the same result may be achieved using `maps:merge/2`
+function (with less room for optimization).
+
+One (probably minor) argument against it is the fact that the "map update"
+syntax is a common reason for very confusing errors programmers make
+when they forget to add a comma between two maps:
+
+    > [#{a => b}
+       #{a => d}].
+    #{a => d}
```
Maria-12648430, line 193: The result is `[#{a => d}]`, not `#{a => d}`.

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+### Map updates via map comprehension
+
+Should map updates be allowed with map-comprehensions? That is should
+this be valid syntax:
+
+    M#{ K => V*2 || K := V <- M}
+
+or this
+
+    M#{ K := V*2 || K := V <- M}
```
Maria-12648430, line 180: The examples, while serving the purpose of illustrating the syntax, are not good ones: The result is the same without the `M` in front.

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+
+There is no final decision about how useful might it be.
```
Maria-12648430, line 182: Personally, I think this would be pretty cool, never thought of it 😅 The accidental "map update" scenario, well, it already exists in the non-comprehension form, so...

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+sometimes will refer to the whole combination of the comprehension, the filter and
+the generator as a "comprehension", because the generator syntax is not
+used outside of the comprehension context.
+
+
+### Syntax
+
+The proposed syntax for the map comprehension part is
+
+    '#{' KeyExpr '=>' ValueExpr '||' GeneratorsFilters '}'
```
Maria-12648430, line 58: There is, well, let's call it an inconvenience, just pointing it out: you can't generate `KeyExpr => ValueExpr` "in one".

For example, while something like this is possible with a list comprehension:
```erlang
%% (double all even numbers and tag as even or odd)
[ if N rem 2 =:= 0 -> {N*2, even}; true -> {N, odd} end || N <- lists:seq(1, 10) ]
```

with map comprehensions, this isn't:
```erlang
#{ if N rem 2 =:= 0 -> N*2 => even; true -> N => odd end || N <- lists:seq(1, 10) }
```
but has to be done with two separate `if`s:
```erlang
#{ if N rem 2 =:= 0 -> N*2; true -> N end => if N rem 2 =:= 0 -> even; true -> odd end; || N <- lists:seq(1, 10) }
```

eeps/eep-0058.md
Maria-12648430: Ok, admittedly, the second one won't work without the first `M` at all =^^=

seriyps: Ah, I just copy-pasted them from the PR comment https://github.com/erlang/otp/pull/4856#issuecomment-853906261
May change to smth like

```erlang
MapToUpdate#{K => V*2 || K := V <- MapToUpdateWith}
```

eeps/eep-0058.md
seriyps: yes, and I've been thinking a lot about this restriction and didn't manage to come with any real solution. Maybe some crazy stuff like defining the comprehension as
```
'#{' <expr> ',' <key expr> => <value expr> '||' ...}
```
so you can write smth like

```
#{
  if N rem 2 =:= 0 -> Key = N*2, Val = even;
     true -> Key = N, Val = odd
  end,
  Key => Val
  || N <- lists:seq(1, 10)
}
```

fenollp: This already is valid:
```erlang
[ {K,V} || N <- lists:seq(1, 10),
       {K,V} <- [if N rem 2 =:= 0 -> {N*2,even}; true -> {N,odd} end]].
```
So this should be as well:
```erlang
#{ K => V || N <- lists:seq(1, 10),
         {K,V} <- [if N rem 2 =:= 0 -> {N*2,even}; true -> {N,odd} end]}
```

seriyps: @fenollp yep, this would work with the spec I'm proposing. But it's really not obvious construction IMO

Maria-12648430: Whoa... I think that looks seriously messy 🤪 I wouldn't really want that, TBH.

Looking at binary comprehensions, the comprehension part must (?) result in a binary/bitstring, for example:
```erlang
> << <<(X+1):8>> || <<X:8>> <= <<"abc">> >>.
<<"bcd">>
```

What if, in a similar manner, the map comprehensions' comprehension part must result in a map instead of a `K => V` construct, which is used to update the already generated map? Aside from solving what I outlined above, this would even allow to insert _multiple_ key-value pairs at a point, as well _not_ inserting anything at all. Again looking at binary comprehensions, both
```erlang
> << <<(X+1):8, (X+2):8)>> || <<X:8>> <= <<"abc">> >>.
<<"bccdde">>
```
and
```erlang
> << <<>> || <<X:8>> <= <<"abc">> >>.
<<>>
```
are allowed.

zuiderkwast: Good point. I don't think it's a problem though. There are always things you can't do and ways to work around it.

Here's a workaround using a "singleton-generator":

```Erlang
[ K => V || N <- lists:seq(1, 10),
            {K, V} <- [if N rem 2 =:= 0 -> {N*2, even}; true -> {N, odd} end] ]
```

eeps/eep-0058.md
zuiderkwast: I think update-comprehensions can be useful, both with `:=` and `=>`.

eeps/eep-0058.md
Maria-12648430: @fennolp @zuiderkwast yes, but as @seriyps pointed out, that's not very obvious. I would call it awkward-hackish even. Aside from that, it requires that secondary helper generator. The compiler might optimize things here, but I doubt it TBH.

I'm not too empathic about this issue actually, I just like to point things out _before_ a ship has sailed ;)

zuiderkwast: There are other things you can't do, such as emitting two elements each round, and workaround using extra generator.

* Example: `[2*N, 2*N + 1 || N <- lists:seq(1, 10)]`
* Workaround: `[X || N <- lists:seq(1, 10), X <- [2*N, 2*N + 1]]`

Maria-12648430: My point, actually: that ship has sailed. And, at a wild guess, not because it was rejected, but because nobody considered it back then. It is possible to do, with a fold or explicit recursion if nothing else, but if we can do it nicely and there is no reason to reject it, why not _do_ it nicely? 😉

seriyps: Oh, I didn't know it's possible with binary comprehensions:
```erlang
> << case (C rem 2) == 0 of true -> <<1>>; false -> <<2>> end || C <- lists:seq(1, 10) >>.
<<2,1,2,1,2,1,2,1,2,1>>
> << case (C rem 2) == 0 of true -> <<1>>; false -> oups end || C <- lists:seq(1, 10) >>. 
** exception error: bad argument
     in function  bit_size/1
        called as bit_size(oups)
```

In that case maybe it makes sense to offer 2 versions of map comprehension: short one
```erlang
'#{' KeyExpr '=>' ValueExpr '||' GeneratorsFilters '}'
```
but then also the one that yields arbitrary map
```erlang
'#{' AnyExprThatReturnsMap '||' GeneratorsFilters '}'
```
it's going to be trickier to optimize though. With short version it's possible to just allocate KeyExpr and ValueExpr to BEAM registers leaving no garbage on heap, but with the 2nd version I guess it would be difficult to avoid creation of temporary maps. I would also be disappointed if programmers start to write
```erlang
#{ #{K => V * 2} || K := V <- Map}
```
just because it's possible or they don't remember what way is correct =)

Maria-12648430: > In that case maybe it makes sense to offer 2 versions of map comprehension

I'm not a huge fan of "there is more than one way to do it", but it might make sense here.

That said, I have only now glimpsed at the code in the reference PR, and while I can't say I grasp it fully, it looks like you turn the map comprehension into a list comprehension, and the resulting `{K, V}` list into a map via `maps:from_list/1`, yes?

In that case, allowing maps as returns would indeed make things difficult. As I see it, instead of a simple `maps:from_list/1` (which is a BIF, btw), you would have to resort to `lists:foldl/3`, like
```erlang
lists:foldl(
    fun
        ({Key, Value}, Acc) -> Acc#{Key => Value};
        (Map, Acc) -> maps:merge(Acc, Map)
    end,
    #{},
    ListCompResult
)
```

seriyps: Our PR was done as a one-day hackathon, so, we have choosen the easiest way it could be done. But eventually it would be much better if we implement a distinct core-erlang code for map comprehensions. Especially given that `{K, V} <- maps:to_list(Map)` would not work for `maps:iterator()`. Also, I'm sure it can be done more efficiently (producing less garbage on heap) if we avoid construction of intermediate tuples and lists in the implementation; but this may be done in a follow-up I think.

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+(because the map generator order is not defined), but is important when
+a map comprehension is used with a list or a binary generator.
+
+### Whether to allow map iterator as input for map generator
+
+Map iterators are allowed in `maps:filter/2`, `maps:fold/3` and
+`maps:map/2`. Since comprehensions are often used in the same
+contexts as those functions, it makes sense to allow map iterators
+so comprehensions could be used as a drop-in replacement for the
+aforementioned functions.
```
seriyps, line 169: I'm still in doubt though... `maps:iterator()` is part of a standard library, but here we making it part of the language? Do we have another examples where this happens?

juhlig: True... I at least don't know of any other place where something like this is done.

Maria-12648430: Neither do I.

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+way as the ordering of the list, resulting from `maps:to_list/1` is
+not guaranteed).
+
+### Behavior when comprehension produced duplicate keys
+
+Same as `maps:from_list/1`: the latest (right-most) value is used
+and the previous values are ignored. It might be less predictable
+when a map comprehension is used in combination with a map generator
+(because the map generator order is not defined), but is important when
+a map comprehension is used with a list or a binary generator.
```
juhlig, line 161: This paragraph is a bit confusing I think, maybe you should phrase it differently?

To be sure, I get the idea: If a map generator is used to generate key-value pairs, and the comprehension produces possibly duplicate keys based on them, it is not predictable which value will be in the resulting map for that key. For example, if a map comprehension is used to flip keys and values in a map, with...
```erlang
#{ V => K || K := V <- #{a => x, b => x} }
```
... the result could be either `#{x => a}` or `#{x => b}`, which one is not predictable (or guaranteed).

With a list (or binary) generator instead, the result is predictable. With...
```erlang
#{ V => K || {K, V} <- [{a, x}, {b, x}] }
```
... the result will _always_ be `#{x => b}`, and with...
```erlang
#{ V => K || {K, V} <- [{b, x}, {a, x}] }
```
... the result will _always_ be `#{x => a}`.

seriyps: Yes, this is exactly what I tried to say.

zuiderkwast: "right-most" isn't really true here. Latest yes. Last generated maybe.

seriyps: I actually copy-pasted this phrase "the latest (right-most) value is used and the previous values are ignored" from OTP documentation of `maps:from_list` http://erlang.org/doc/man/maps.html#from_list-1

eeps/eep-0058.md
```diff
@@ -0,0 +1,255 @@
...
+========
+
+This EEP proposes syntax and semantics for map comprehensions and
+generators, similar to list and binary comprehensions and generators,
+but operating on maps and map iterators.
+
+Copyright
+=========
+
+This document has been placed in the public domain.
```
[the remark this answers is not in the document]
seriyps: Updated, thanks

seriyps:
Hey, is there any chance this get any traction? I'm afraid there is not so much time left to squeeze this feature in OTP-25

KennethL:
This feature will not make it into OTP 25. The OTP team will take a deeper look after the OTP 25 release (May 2022) and decide if it can be introduced in OTP 26.

PR #35, opened 2021-11-02
Update eep-0049 with decisions from OTP Technical Board

bjorng:
Introduce `maybe` as keyword to start the construct and `<~' as matching operator.

### DIFF
--- a/eeps/eep-0049.md
+++ b/eeps/eep-0049.md
@@ -10,13 +10,11 @@ EEP 49: Value-Based Error Handling Mechanisms
 Abstract
 ========
 
-This EEP extends the `begin ... end` expression to make it a construct
-usable for control flow and value-based error handling based on pattern
-matching.
-
-This introduces `begin ... else ... end` along with a new contextual use
-of the `<-` operator to replace or simplify deeply-nested `case ... end`
-expressions, and prevent using exceptions for control flow.
+This EEP introduces the `maybe ... end` expression as a construct
+usable for control flow and value-based error handling based on
+pattern matching. By using this construct, deeply nested `case
+... end` expressions can be avoided or simplified, and the use of
+exception for flow control can be avoided.
 
 Copyright
 =========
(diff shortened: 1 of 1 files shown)


eeps/eep-0049.md
```diff
@@ -160,9 +157,9 @@ construct:
 
     commit_write(OpaqueData) ->
         begin
```
zuiderkwast, line 159:
```suggestion
        maybe
```

eeps/eep-0049.md
```diff
@@ -171,9 +168,9 @@ Or, to protect against `disk_log` calls returning something else than `ok |
...
```
zuiderkwast, line 170:
```suggestion
        maybe
```

eeps/eep-0049.md
```diff
@@ -221,24 +218,25 @@ as:
 
     pre_check(Action, User, Context, ExternalThingy) ->
         begin
```
zuiderkwast, line 220:
```suggestion
        maybe
```

eeps/eep-0049.md
```diff
@@ -1399,38 +1402,30 @@ is chosen following Erlang/OTP standards:
...
-compatibility concern to be had.
-
-The need for the `<-` operator to be used in a new context means code built
-with support for the new expressions won't be portable to older Erlang
-releases. However, if all the code is rewritten _after_ the AST and reuses
-existing core Erlang components, built BEAM artifacts should work on older
-versions fine. This is, however, not a supported use case by the OTP team.
+The plan is to introduce this new feature together with a mechanism
+that let the user activate it and other new features individually per
+module and by that the new keyword `maybe` will only be a potential
```
zuiderkwast, line 1418: Will there be another EEP for this mechanism?

Will a feature become activated-by-default after some number of OTP versions?

zuiderkwast (commented):
I'm not sure whether I like it or not, but I found some typos. :-)

ferd (approved):

eeps/eep-0049.md
bjorng: > Will there be another EEP for this mechanism?

Yes, there is ongoing work on both an EEP and an implementation. We are still doing internal reviews of the EEP.

> Will a feature become activated-by-default after some number of OTP versions?

If a feature turns out to be successful, we will activate it by default in a future release (and there might be a possibility that the feature could still be possible to be turned off for legacy code).

PR #39, opened 2022-02-28
Add EEP for experimental features

lisztspace:
Add EEP describing support for experimental features in Erlang/OTP

### DIFF
--- a/eeps/eep-0060.md
+++ b/eeps/eep-0060.md
@@ -1,0 +1,685 @@
+    Author: Cons T Åhs <cons(at)erlang(dot)org>
+    Status: Draft
+    Type: Standards Track
+    Erlang-Version: OTP25
+    Created: 4-Oct-2021
+    Post-History:
+****
+EEP 60: Introduce Support for Experimental Features
+----
+
+Abstract
+========
+
+This EEP provides a suggestion on how to allow support for enabling
+and disabling language features.  This will, e.g., allow users to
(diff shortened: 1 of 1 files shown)


ferd:
This has been merged fast, but I would recommend putting in more weight towards the configuration formats that can be represented as `{enable_feature, true | false, Name}` or some variation thereof because those can be put in a proplist (or a map) and automatically allow to set or override preferences by following the specification order.

By comparison, code that can use both `{enable, X}` or `{disable, X}` and multiple variations will instead require pre-/post-processing by tools to properly allow mergeable configuration formats.

lisztspace:
My bad, it shouldn't have been merged directly.

I agree that changing the internal format to something like you suggest is a good idea and will give it some thought.  For `erlc`, using `-enable-feature` and `-disable-feature` might be preferred for readability.  I already have some code that uses the ordering, of course.

ferd:
Yeah, I think that's fine, I'm more concerned about the format passed to the compiler module, which I assume the CLI can translate however it wants. Transparently, this is because of challenges in making a more general option-merging mechanism for rebar3 profiles with `erl_opts` configuration values. But I'm sure that in general, other tools may end up with similar issues in the long run.

lisztspace:
Note: I've created a new PR (https://github.com/erlang/eep/pull/40) to for better discussions on this.

We have internally discussed using `{feature, enable | disable, name}` as the compiler option since it aligns better with the module directive `-feature(enable | disable, name).`

PR #40, opened 2022-03-01
New EEP: add support for experimental features

lisztspace:
New EEP: add support for experimental features

(take 2 to revert premature merge)

### DIFF
--- a/eeps/eep-0060.md
+++ b/eeps/eep-0060.md
@@ -1,0 +1,685 @@
+    Author: Cons T Åhs <cons(at)erlang(dot)org>
+    Status: Draft
+    Type: Standards Track
+    Erlang-Version: OTP25
+    Created: 4-Oct-2021
+    Post-History:
+****
+EEP 60: Introduce Support for Experimental Features
+----
+
+Abstract
+========
+
+This EEP provides a suggestion on how to allow support for enabling
+and disabling language features.  This will, e.g., allow users to try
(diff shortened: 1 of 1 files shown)


ferd:
Porting my comments from the prior PR:

 I would recommend putting in more weight towards the configuration formats that can be represented as `{enable_feature, true | false, Name}` or some variation thereof because those can be put in a proplist (or a map) and automatically allow to set or override preferences by following the specification order.

I am not concerned about the command-line options, more about the format passed to the compiler module, which I assume the CLI can translate however it wants. Transparently, this is because of challenges in making a more general option-merging mechanism for rebar3 profiles with `erl_opts` configuration values. But I'm sure that in general, other tools may end up with similar issues in the long run. 

Regarding the prior PR's comment mentioning `{feature, enable | disable, name}`, I do like that form better, but I may suggest using `{feature, Name, enable | disable}` -- for the simple reason that you can sort it in tuple order on the first 2 elements and assume the third value is stable (the option's value), as is the case with `{d, Macro, Value}`. As a data list, it feels more natural to have a general `{Type, Key, Val}` ordering, at least to me.

lisztspace:
As said in my my from the prior PR, the alternative format `{feature, enable | disable, name }` has been discussed.  Your suggestion of swapping argument order to `{feature, name, enable | disable}` and the motivation is excellent for all the reasons you list.  It is also reasonable to change in the same way for the `-feature(..)` directive.

PR #47, opened 2023-06-08
EEP 64: Triple-Quoted Strings

kikofernandez:
This EEP discusses the design of the triple-quoted binary strings.

In the case that document attributes [EEP 59](https://www.erlang.org/eeps/eep-0059) and interpolation strings ([EEP 62](https://github.com/erlang/eep/pull/45)) are added before this EEP, interpolation attributes are to be disallowed in documentation attributes (this is mentioned in the EEP).

The semantics should feel familiar for Elixir developers, and should be pretty close to Elixir triple-quote semantics.

Feedback is welcome.

### DIFF
--- a/eeps/eep-0064.md
+++ b/eeps/eep-0064.md
@@ -186,7 +186,7 @@ the string's content:
 Note that the following could be a syntax error; too short
 multi-line string since trailing newline shall be stripped
 both from the starting line and from the last content line,
-so the content could be seen as less then empty.
+so the content could be seen as less than empty.
 But it is more convenient to allow it as also an empty string:
 
     "" = """
@@ -285,8 +285,8 @@ would be syntax errors.
   syntax errors.  Only a few will have a subtly changed
   behaviour (string content).
 * Users can simply grep for `"""` in their source code.
-  Causing the same sequence e.g through macros would
-  be harder to find, but the worst problem would not
+  Creating the same sequence e.g through macros would
+  be harder to find; the worst problem would not
   be new syntax errors (hard to miss), but changed
   behaviour.  And the changed behaviour would be
   a slightly different string content.
--- a/eeps/eep-0064.md
+++ b/eeps/eep-0064.md
@@ -155,12 +155,12 @@ value term.  It starts and ends with three double quotes: `"""`.
 Double quotes, `"`, are chosen because normal Erlang strings use them
 and this is just a new variant.  Since double quotes are used
 a triple-quoted string shall, as a normal string,
-produce a list of code-points.
+produce a list of characters (Unicode code-points).
 
 It would be more convenient if a triple-quoted string produced
 an UTF-8 binary, but that would be a surprising feature for double quotes,
 and the documentation build process can work around this by converting
-the code-point list (`string()`) into the needed binary chunk.
+the character list into the needed binary chunk.
 
 In source code a triple-quoted string is valid within a binary,
 so producing a Unicode binary is reasonably straightforward:
@@ -176,7 +176,7 @@ multi-line strings here.
 As a future expansion it has been proposed to use [Sigils][] (prefixes)
 for specialized strings such as regular expressions,
 interpolated variables ([PR-7343][]), Unicode binary strings, etc.
-For example: `X = ~u"Tschüß"` for an UTF-8 encoded string.
+For example: `X = ~u"Tschüß"` for an UTF-8 encoded binary.
 
 ### Triple-Quoted String Start
 
--- a/eeps/eep-0064.md
+++ b/eeps/eep-0064.md
@@ -6,7 +6,7 @@
     Erlang-Version: OTP-27
     Post-History:
 ****
-EEP 64: Multi-line Indented Verabatim Strings
+EEP 64: Multi-line Indented Verbatim Strings
 ----
 
 Abstract
@@ -101,7 +101,7 @@ the documentation build process can convert the codepoint list
 into the needed binary chunk.
 
 In source code a MIV String is valid in a binary so
-producing a unicode binary is reasonably straightforward:
+producing a Unicode binary is reasonably straightforward:
 
     X = <<"""
         Line 1
@@ -113,12 +113,12 @@ multi-line strings here.
 
 As a future expansion it has been proposed to use prefixes
 for specialized strings such as regular expressions,
-interpolated variables ([PR-7343][]), unicode binary strings, etc.
+interpolated variables ([PR-7343][]), Unicode binary strings, etc.
 For example: `X = u"Tschüß"` for an UTF-8 encoded string.
 
 ### MIV String Start
 
-After the starting `"""` only whitespace is allowed
+After the starting `"""` only white-space is allowed
 up to the end of the line.
 
 As a possible future expansion we might allow a keyword here
@@ -138,7 +138,7 @@ A later step, preferably done by the parser, strips the
 characters up to and including the newline from the
 first line of the string.
 
-If any of these characters is not whitespace,
+If any of these characters is not white-space,
 the parser reports a syntax error.
 
 ### MIV String End
@@ -146,21 +146,21 @@ the parser reports a syntax error.
 All bytes are collected as they are (verbatim) and becomes
 the MIV String content.
 
-A MIV String ends with whitespace on a new line followed by `"""`.
+A MIV String ends with white-space on a new line followed by `"""`.
 This completes the scanner token.
 
-A later step, preferably done by the parser, uses the whitespace
+A later step, preferably done by the parser, uses the white-space
 on the ending line as the definition of the string's indentation
-and strips that whitespace seqence from every line in the string,
+and strips that white-space sequence from every line in the string,
 and strips the newline preceding the ending line.
 
 If any of the lines should not start with the defined indentation
 either because the line is too short or if the prefix differs,
 the parser reports a syntax error.
 
-Requiring that all lines must have exactly the same whitspace
+Requiring that all lines must have exactly the same white-space
 characters as indentation is a simple solution to not have
-to define how indentation whitespace (tab vs. space) normalization
+to define how indentation white-space (tab vs. space) normalization
 should be done, and also seems like a reasonable requirement.
 
 ### Leading and trailing newline
@@ -182,13 +182,13 @@ the string's content:
     "" = """
          """
 
-The definition of newline and whitespace is the same as
+The definition of newline and white-space is the same as
 the current in the scanner, but; when stripping the newline
 from the line preceding the ending line, a CR should also
 be stripped, if the line ends in CR LF.
 
 This is a convenience for systems where CR LF is used as newline.
-In most places in the scanner CR is treated as whitespace,
+In most places in the scanner CR is treated as white-space,
 but in this case it would be inconvenient to not strip the CR.
 
 ### Indentation
@@ -218,7 +218,7 @@ determines the indentation.
         """,
 
     """ This is a syntax error
-    (non-whitespace on start line)
+    (non-white-space on start line)
     """,
 
     """
@@ -262,7 +262,7 @@ Also, this is valid today:
 
 But according to this EEP it would be two syntax errors:
 
-1. The start line has got non-whitespace after `"""`.
+1. The start line has got non-white-space after `"""`.
 2. The first content line has incorrect indentation.
 
 There are many other similar constructions that also
@@ -281,7 +281,7 @@ would be syntax errors.
   behaviour.  And the changed behaviour would be
   a slightly different string content.
 
-Therefore, it should be very unlikly that anyone
+Therefore, it should be very unlikely that anyone
 encounters a real backwards incompatibility problem
 from the suggestions in this EEP.
 
--- a/eeps/eep-0064.md
+++ b/eeps/eep-0064.md
@@ -170,8 +170,8 @@ so producing a Unicode binary is reasonably straightforward:
         Line 2
         """/utf8>>
 
-The 2 + 7 characters overhead is not exhausting since we are targeting
-multi-line strings here.
+The 2 + 7 characters ("`<<`" + "`/utf8>>`") overhead is not exhausting
+since we are targeting multi-line strings here.
 
 As a future expansion it has been proposed to use [Sigils][] (prefixes)
 for specialized strings such as regular expressions,
--- a/eeps/eep-0064.md
+++ b/eeps/eep-0064.md
@@ -197,11 +197,11 @@ The scanner does not need to have any special treatment
 for the string content on the line after the starting
 `"""` except that it should not search for an ending `"""`.
 
-A later step, preferably done by the parser, strips the
-characters up to and including the newline from the string content.
+A later step strips the characters up to and including the newline
+from the string content.
 
 If any of these characters is not white-space,
-the parser reports a syntax error.
+   a syntax error is reported.
 
 ### Triple-Quoted String End
 
@@ -211,14 +211,14 @@ the string content.
 A triple-quoted string ends with newline followed by optional white-space
 and then `"""`.  This completes the scanner token.
 
-A later step, preferably done by the parser, uses the white-space
-on the ending line as the definition of the string's indentation
+A later step uses the white-space on the ending line
+as the definition of the string's indentation
 and strips that particular white-space sequence from every line
 in the string, and strips the newline preceding the ending line.
 
 If any of the lines do not start with the defined indentation
 either because the line is too short or if the prefix differs,
-the parser reports a syntax error.
+a syntax error is reported.
 
 Requiring that all lines must have exactly the same indentation
 characters is a simple solution to not have to define how indentation
--- a/eeps/eep-0064.md
+++ b/eeps/eep-0064.md
@@ -1,175 +1,342 @@
-    Author: Kiko Fernandez-Reyes <kiko(at)erlang(dot)org>
+    Author: Kiko Fernandez-Reyes <kiko(at)erlang(dot)org>,
+        Raimo Niskanen <raimo(at)erlang(dot)org>
     Status: Draft
     Type: Standards Track
     Created: 07-Jun-2023
     Erlang-Version: OTP-27
     Post-History:
 ****
-EEP XXX: Triple-quoted binary strings
+EEP 64: Multi-line Indented Verabatim Strings
 ----
 
 Abstract
 ========
 
-This EEP proposes the introduction of triple-quoted binary-strings syntax and
-runtime semantics. The main benefit is to allow multi-line binary strings in an
-easy way, similar to other languages, e.g., Elixir.
+This EEP proposes the introduction of Multi-line Indented Verbatim strings,
+*MIV Strings* and defines their semantics.  The main benefit is to allow
+*multi-line* strings in an easy and useful, *indented*, way,
+similar to in other languages, e.g Elixir.
+
+Their first use case is for in-module documentation attributes
+containing [Markdown][] or similarly formatted text where *verbatim* text
+is desirable since any documentation text format has its own notion
+of escape sequences which will collide with Erlang's escape sequences.
 
 Rationale
 =========
 
-One limitation of Erlang today (June 2023) is the ability to write multi-line
-string binaries. That said, the main reason to consider this EEP is
-[EEP-59][] where documentation attributes can
-benefit from multi-line string binaries: the documentation generates Docs chunks
-format -- documentation in its binary format-- which saves space and makes the
-documentation available to the shell.
+Today (June 2023), writing multi-line strings is awkward and arguably ugly.
+They may contain escape sequences and have no concept of indentation:
 
-Triple-Quoted Binary-String Design Decisions
----------------------------------------------
+    foo () ->
+        case bar() of
+             ok ->
+                 X = "First line
+    Second line with \"\\*not emphasized\\* Markdown\"
+    Third line",
+                 {ok, X}
+        end.
 
-### Runtime semantics
+The content's indentation cannot adhere to the surrounding code's
+and the `*` has to be doubly escaped to get a  `\*` character
+sequence into the actual content.
 
-Triple-quoted strings should only produce binary-strings. This makes easy to
-avoid typing issues with union types, and makes clear that a triple-quoted string
-does not need function with guards to test if the produce result is a list or a
-binary (this can happen if Erlang adds string interpolation,
-[EEP-62][],
-[PR-7343][]).
+In a documentation attribute as suggested in [EEP 59][]
+the indentation problem is not that pronounced because
+the documentation attribute itself is not much indented:
 
-Triple-quoted binary-strings do not allow string interpolation and, if we were
-to accept them, only calls to statically known values should be allowed, e.g.,
-macros. This design decision forbids the creation of dynamic documentation that
-generates code at runtime, e.g., doing an IO call to get a value not known
-statically and based on it perform an action (notice that this is allowed when
-using string interpolation).
+    -doc """
+        First line
+        Second line with "\*not emphasized\* Markdown"
+        Third line
+        """.
 
-### Binary-String Design
+The main reason to consider this EEP is for documentation
+attributes, and here not having to worry about escape sequences
+is the most attractive property of this EEP.  But introducing
+a new string format will also require defining how it would behave
+in Erlang code.
 
-To make the usage of binary-string similar to other languages, this EEP proposal
-makes the binary-string design decisions similar to the ones for the Elixir language.
+Having a string format only allowed in attributes would simply be
+very strange and the one suggested in this EEP would be useful
+also in Erlang code.
 
-#### Binary-String Start and End
+Design Decisions
+----------------
 
-Binary-strings must start and end with triple-quotes in their own lines:
+An attribute is an Erlang form in the source code
+that consists of a `-` token, an atom, one value term and a full stop
+(dot).  The value term may be enclosed in parentheses
+(which is not very interesting for documentation attributes).
 
-*Code Example*
+    -doc "  Badly formatted
+    documentation paragraph
+    /-\\
+    \\-/".
+
+A documentation attribute should have a string as its content term,
+and here we want to use our new and more convenient MIV String
+instead of a normal string:
+
+    -doc """
+          Better formatted
+        documentation paragraph
+        /-\
+        \-/
+        """.
+
+### MIV String Scanner Token
+
+A MIV String must be a token that the scanner recognizes as a string,
+which makes it suitable for a documentation attribute value term.
+It starts and ends with three double quotes: `"""`.
+
+Double quotes, `"`, are chosen because normal Erlang strings use them
+and this is just a new variant.  Since double quotes are used
+a MIV String should, as a normal string, produce a list of codepoints.
+
+It would be more convenient if a MIV String produced an UTF-8 binary,
+but that would be surprising when using double quotes, and
+the documentation build process can convert the codepoint list
+into the needed binary chunk.
+
+In source code a MIV String is valid in a binary so
+producing a unicode binary is reasonably straightforward:
+
+    X = <<"""
+        Line 1
+        Line 2
+        """/utf8>>
+
+The extra overhead is not exhausting since this we are targeting
+multi-line strings here.
+
+As a future expansion it has been proposed to use prefixes
+for specialized strings such as regular expressions,
+interpolated variables ([PR-7343][]), unicode binary strings, etc.
+For example: `X = u"Tschüß"` for an UTF-8 encoded string.
+
+### MIV String Start
+
+After the starting `"""` only whitespace is allowed
+up to the end of the line.
+
+As a possible future expansion we might allow a keyword here
+that shouldn't be part of the string content, but could be
+a hint for for syntax highlighting in the editor.
+
+    -doc """ md
+        Markdown content
+        * Bullet list
+        """.
+
+The scanner does not need to have any special treatment
+for the string content on the line after the starting
+`"""` except that it should not search for an ending `"""`.
+
+A later step, preferably done by the parser, strips the
+characters up to and including the newline from the
+first line of the string.
+
+If any of these characters is not whitespace,
+the parser reports a syntax error.
+
+### MIV String End
+
+All bytes are collected as they are (verbatim) and becomes
+the MIV String content.
+
+A MIV String ends with whitespace on a new line followed by `"""`.
+This completes the scanner token.
+
+A later step, preferably done by the parser, uses the whitespace
+on the ending line as the definition of the string's indentation
+and strips that whitespace seqence from every line in the string,
+and strips the newline preceding the ending line.
+
+If any of the lines should not start with the defined indentation
+either because the line is too short or if the prefix differs,
+the parser reports a syntax error.
+
+Requiring that all lines must have exactly the same whitspace
+characters as indentation is a simple solution to not have
+to define how indentation whitespace (tab vs. space) normalization
+should be done, and also seems like a reasonable requirement.
+
+### Leading and trailing newline
+
+The rules above strips one leading and one trailing newline.
+This is a simple convention that also gives control over
+the string's content:
+
+    "\n  X\n" = """
+        
+          X
+        
+        """,
+    
+    "X" = """
+        X
+        """,
+    
+    "" = """
+         """
+
+The definition of newline and whitespace is the same as
+the current in the scanner, but; when stripping the newline
+from the line preceding the ending line, a CR should also
+be stripped, if the line ends in CR LF.
+
+This is a convenience for systems where CR LF is used as newline.
+In most places in the scanner CR is treated as whitespace,
+but in this case it would be inconvenient to not strip the CR.
+
+### Indentation
+
+The rules above facilitates indentation of the content
+to adhere to the surrounding code.  The ending line
+determines the indentation.
+
+    "This text\nhas no indentation" = """
+        This text
+        has no indentation
+        """,
+    
+    "    This text\n    has indentation" = """
+            This text
+            has indentation
+        """,
+    
+    "  This text\nhas an indented first line" =
+        """
+          This text
+        has an indented first line
+        """,
+    
+    """
+    This is a syntax error (incorrect indentation)
+        """,
+
+    """ This is a syntax error
+    (non-whitespace on start line)
+    """,
+
+    """
+    This will probably be a syntax error
+    since no ending line can be found"""
+
+### Backwards incompatibility
+
+This is valid today:
 
     X = """
-        This is the beginning of the triple-quote text
+        X
         """
 
-*Documentation Example*
+It is equivalent to:
+
+    X = "" "
+        X
+        " ""
+
+Which is equivalent to:
+
+    X = "
+        X
+        "
+
+Which is equivalent to:
+
+    X = "\n    X\n"
+
+But with the suggested MIV Strings the first
+code snippet would instead be equivalent to:
+
+    X = "X"
+
+Also, this is valid today:
+
+    X = """ xxx
+      X
+        """
+
+But according to this EEP it would be two syntax errors:
+
+1. The start line has got non-whitespace after `"""`.
+2. The first content line has incorrect indentation.
+
+There are many other similar constructions that also
+would be syntax errors.
+
+* It is far from likely that anyone has deliberately
+  used `"""` in source code to mean an empty string
+  concatenated to another string.
+* Most today allowed combinations with `"""` will cause
+  syntax errors.  Only a few will have a subtly changed
+  behaviour (string content).
+* Users can simply grep for `"""` in their source code.
+  Causing the same sequence e.g through macros would
+  be harder to find, but the worst problem would not
+  be new syntax errors (hard to miss), but changed
+  behaviour.  And the changed behaviour would be
+  a slightly different string content.
+
+Therefore, it should be very unlikly that anyone
+encounters a real backwards incompatibility problem
+from the suggestions in this EEP.
+
+### Quoting of `"""`
+
+In the rules above there is no possibility to have `"""`
+first on a line in a MIV String.
+
+This would be allowed:
 
     -doc """
-         Removes double-quotes (") from a given string
-         """
-    remove_double_quotes(X) -> 
+        A MIV String starts with: """
+        and ends with: """
+        """.
 
-#### Binary-Strings Closing of Triple-Quotes
+As long as `"""` isn't first on a line.
 
-The closing of the triple-quotes denotes the space (indentation):
+It would be possible to work around in Erlang code:
 
-*1. Code Example*
+    X = """
+        A MIV String starts with:
+        ""
+        """ """
+        "
+        and ends with:
+        ""
+        """ "\""
 
-    >> X = """
-           This text
-           has no indentation
-           """
-    
-    Equivalent to: <<"This text\nhas no indentation\n">>
+That is ugly, and it is not possible in a documentation
+attribute where string concatenation isn't allowed.
 
-*2. Code Example*
+We can either ignore the problem since it is only
+when placed first on a line that `"""` is a problem,
+or we can use the GitHub [Markdown][] trick to allow
+3 or more start characters and matching end characters
+so this would be valid:
 
-    >> Y = """
-           This text
-           has indentation
-    """
-    
-    Equivalent to: <<"       This text\n       has indentation\n">>
-
-*3. Code Example*
-
-    >> Z = """
-      This text has
-      two space indentation
-    """
-    
-    Equivalent to: <<"  This text has\n  two space indentation\n">>
-
-*1. Documentation Example - No indentation*
-
-    -doc """
-         Removes double-quotes (") from a given string
-         """
-    remove_double_quotes(X) -> 
-
-*2. Documentation Example - No indentation*
-
-    -doc """
-    Removes double-quotes (") from a given string
-    """
-    remove_double_quotes(X) -> 
-
-*3. Documentation Example - Indentation*
-
-    -doc """
-         Removes double-quotes (") from a given string
-    """
-    remove_double_quotes(X) -> 
-
-#### Binary-Strings Errors
-
-An error should be raised when there is text that starts before the closing of triple-quotes, where
-the error makes the code written using triple-quotes to have a uniform style:
-
-*Code Example*
-
-    >> Err = """
-    This shall be an error
-             """
-   
-    error             
-
-*Documentation Example*
-
-    -doc """
-    Removes double-quotes (") from a given string
-         """
-    remove_double_quotes(X) -> 
-
-#### Escaping Newlines in Triple-Quotes
-
-Escaping new lines using `\`:
-
-To write `\` inside triple quotes, one needs to use `\\`. [EEP-59][] should
-ignore the meaning of `\` if inside a code block.
-
-*Code Example*
-
-    >> X = """
-           This text \
-           has no indentation
-           """
-    
-    Equivalent to: <<"This text has no indentation\n">>
-
-*Documentation Example*
-
-    -doc """
-    Removes double-quotes (") \
-    from a given string
-    """
-    remove_double_quotes(X) -> 
+    X = """"
+        A MIV String starts with:
+        """
+        and ends with:
+        """
+        """"
 
 [EEP 59]: https://www.erlang.org/eeps/eep-0059
     "EEP 59: Module attributes for documentation"
 
-[EEP-62]: https://www.erlang.org/eeps/eep-0062
+[EEP 62]: https://www.erlang.org/eeps/eep-0062
     "String Interpolation Syntax"
 
 [PR-7343]: https://github.com/erlang/otp/pull/7343
     "Feature: String Interpolation"
+
+[Markdown]: https://github.github.com/gfm/
+    "GitHub Flavored Markdown"
 
 Copyright
 =========
--- a/eeps/eep-0064.md
+++ b/eeps/eep-0064.md
@@ -1,0 +1,187 @@
+    Author: Kiko Fernandez-Reyes <kiko(at)erlang(dot)org>
+    Status: Draft
+    Type: Standards Track
+    Created: 07-Jun-2023
+    Erlang-Version: OTP-27
+    Post-History:
+****
+EEP XXX: Triple-quoted binary strings
+----
+
+Abstract
+========
+
+This EEP proposes the introduction of triple-quoted binary-strings syntax and
+runtime semantics. The main benefit is to allow multi-line binary strings in an
+easy way, similar to other languages, e.g., Elixir.
+
+Rationale
+=========
+
+One limitation of Erlang today (June 2023) is the ability to write multi-line
+string binaries. That said, the main reason to consider this EEP is
+[EEP-59][] where documentation attributes can
+benefit from multi-line string binaries: the documentation generates Docs chunks
+format -- documentation in its binary format-- which saves space and makes the
+documentation available to the shell.
+
+Triple-Quoted Binary-String Design Decisions
+---------------------------------------------
+
+### Runtime semantics
+
+Triple-quoted strings should only produce binary-strings. This makes easy to
+avoid typing issues with union types, and makes clear that a triple-quoted string
+does not need function with guards to test if the produce result is a list or a
+binary (this can happen if Erlang adds string interpolation,
+[EEP-62][],
+[PR-7343][]).
+
+Triple-quoted binary-strings do not allow string interpolation and, if we were
+to accept them, only calls to statically known values should be allowed, e.g.,
+macros. This design decision forbids the creation of dynamic documentation that
+generates code at runtime, e.g., doing an IO call to get a value not known
+statically and based on it perform an action (notice that this is allowed when
+using string interpolation).
+
+### Binary-String Design
+
+To make the usage of binary-string similar to other languages, this EEP proposal
+makes the binary-string design decisions similar to the ones for the Elixir language.
+
+#### Binary-String Start and End
+
+Binary-strings must start and end with triple-quotes in their own lines:
+
+*Code Example*
+
+    X = """
+        This is the beginning of the triple-quote text
+        """
+
+*Documentation Example*
+
+    -doc """
+         Removes double-quotes (") from a given string
+         """
+    remove_double_quotes(X) -> 
+
+#### Binary-Strings Closing of Triple-Quotes
+
+The closing of the triple-quotes denotes the space (indentation):
+
+*1. Code Example*
+
+    >> X = """
+           This text
+           has no indentation
+           """
+    
+    Equivalent to: <<"This text\nhas no indentation\n">>
+
+*2. Code Example*
+
+    >> Y = """
+           This text
+           has indentation
+    """
+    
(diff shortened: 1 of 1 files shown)


eeps/eep-0064.md
```diff
@@ -0,0 +1,187 @@
...
+    remove_double_quotes(X) -> 
+
+*3. Documentation Example - Indentation*
+
+    -doc """
+         Removes double-quotes (") from a given string
+    """
+    remove_double_quotes(X) -> 
+
+#### Binary-Strings Errors
```
josevalim, line 121: Perhaps worth mentioning it also errors when there is a `"""` and it is not immediately followed spaces and a newline? Elixir says:

```
iex(1)> """foo"""
** (SyntaxError) iex:1:1: heredoc allows only optional whitespace followed by a new line after """
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,187 @@
...
+    "EEP 59: Module attributes for documentation"
+
+[EEP-62]: https://www.erlang.org/eeps/eep-0062
+    "String Interpolation Syntax"
+
+[PR-7343]: https://github.com/erlang/otp/pull/7343
+    "Feature: String Interpolation"
+
+Copyright
+=========
```
josevalim, line 175: Should there be a discussion about compatibility? In particular, `"""foo"""` is valid Erlang today. It is very unlikely, but perhaps worth mentioning.

eeps/eep-0064.md
```diff
@@ -0,0 +1,187 @@
...
+benefit from multi-line string binaries: the documentation generates Docs chunks
+format -- documentation in its binary format-- which saves space and makes the
+documentation available to the shell.
+
+Triple-Quoted Binary-String Design Decisions
+---------------------------------------------
+
+### Runtime semantics
+
+Triple-quoted strings should only produce binary-strings. This makes easy to
```
josevalim, line 33: I worry this may be potentially confusing. Why is `"foo"` a list but `"""foo"""` a binary? What if I want to write a long text but as a charlist instead? Here is [a potential example](https://github.com/elixir-lang/elixir/blob/0872ee5444f257744866a4d4c237057f2d97d601/lib/elixir/src/elixir_parser.yrl#L1237-L1249). So, even though I would prefer binaries, I believe a more consistent option is to return charlists.

Then, for binaries, there are a few options:

1. Write `<<"""...""">>`

2. Build on EEP 63 to introduce `u"""..."""`

3. However, for documentation in particular, the compiler can convert `-doc """...""".` into binaries, so the distinction may not be terribly important

asabil: Completely agreeing with José here, I think triple quotes strings should produce strings in erlang unless I closed in a binary context, this would be more consistent with the existing Erlang syntax.

essen:
I'm not a fan of triple quoting for this. Providing an end marker avoids having to worry about indentation and whatnot. Something like this in a terminal:

```
<<EOF
hello
EOF
```

But of course adapted to Erlang syntax. The end marker is whatever you need, it could be `<<PIE` or other.

eeps/eep-0064.md
garazdawi: I agree with José as well, I think it is confusing to use `""""` as a binary. Another option for unicode binaries could be to use backticks:

    io:format(`backtick unicode binary`)
    -doc ```
           backtick 
           multiline
           unicode
           binary
         ```.
Not the most convenient key on some keyboards, but Javascript seems to get away with using it so maybe we can aswell.

bjorng: I like using backticks to mean binary. That nicely solves how to write binary literal strings on one line as well. I think we have a winner!

zuiderkwast: Backticks is interesting because it also exists in Prolog. Erlang's syntax comes from Prolog as we all know (which also has single quote for atoms). In SWI Prolog it seems to be configurable how double-quoted and back-quoted strings work: https://www.swi-prolog.org/pldoc/man?section=string and in [GNU Prolog](http://gprolog.org/manual/gprolog.html#set-prolog-flag%2F2) similarily. 

But, for simple binary strings, isn't a prefix like `b"binary string"` simpler and more aligned with Python, C, Rust, etc. for different kinds of strings?

I think `b""` is enough for UTF-8 encoded binaries as long as the source code is UTF-8 encoded, which it normally is. (If we then add triple quotes for charlist strings, then `b"""..."""` seems the natural choice for the binary version.)

(Btw [C11](https://en.cppreference.com/w/c/language/string_literal) uses `u""` for UTF-16 strings, `U""` for UTF-32 strings and `u8""` for UTF-8, so `u""` for UTF-8 would be confusing in this context.)

josevalim: There are currently three proposals for revamping strings in Erlang:

* Adding string interpolation
* Adding a shortcut notation for utf-8 encoded binaries
* Adding support for heredocs/triple-quoted strings

The tricky part is: we cannot consider each of them in isolation. `"charlists"` cannot receive string interpolation today because that would be backwards incompatible. If we introduce `` `utf-8 encoded binaries` ``, if they are to receive interpolation, the interpolation syntax must be reserved now. At the same time, I feel it would be inconsistent if utf-8 encoded binaries (backticks) get interpolation and strings (double-quoted) do not, so I feel like choosing backticks would close the door on string interpolations. Which is a fine decision to make as long as we are intentional about it.

The argument for `u"..."` style is that it provides a general syntax. For example, regexes are tricky to write in Erlang because you need to consistently double escape. A `r"..."` syntax would solve this problem (and we could precompile regexes too, instead of calling `re:compile/1`). We could even make it user-extensible in the future, similar to how it works in JavaScript, allowing a developer to have their own `merl"..."` or `xml"..."` strings.

I am not necessarily arguing in favor of all of these extensions but my point is that `xyz"..."` provides room for growth. Even if we say: we don't want to have string interpolation now, will that still be true in 5 years or 10 years? The upcoming Java 21 is getting string interpolation.

Finally, choosing backtick will also depend on the documentation format. If we choose Markdown, the backticks in Markdown are used for code snippets. Which means that, if I want to write "string:to_upper(`hello`)" in the docs, I will have to surround it by double backticks, like this: ``` ``string:to_upper(`hello`).`` ``` and I am afraid this will frequently trip people up.

---

Overall, my vote would be the following:

1. Introduce triple-quoted strings as in this proposal but have them return lists (this means we don't care about binaries for now)

2. Because `"foo"` does not have interpolation, it is expected `"""foo"""` does not have interpolation either (this means we don't care about interpolation for now)

3. Allow `-doc` to receive either binary strings or list strings and document they are normalized to binary strings

This allows us to side-step the other proposals for now. Then we should discuss if-when-how to add interpolation and if-when-how to add a syntax shortcut for binaries.

eeps/eep-0064.md
kikofernandez: Since we are still compatible with existing code, and one can write strings using single quotes, we may have to allow code that for some reason is `"""foo"""`, which in Erlang produces `"foo"`. If we use backtick instead of quotes, then your point stands

josevalim: Perhaps it is worth pushing a deprecation warning to Erlang/OTP 26.x that warns on triple quotes?

eeps/eep-0064.md
RaimoNiskanen: We can introduce interpolation in all strings, by using an improbable escape sequence: `\(`...`)` in any string.
Then would choosing backticks for binary strings not close the door for interpolation.

ferd: As a side note, that improbable string sequence specifically (`\(...)`) would absolutely risk clashing with a bunch of regexes though, as a way to match on a parens without creating a group.

RaimoNiskanen: Yes, but a regexp string is a different animal where I guess you do not want any escape sequences, and therefore cannot use interpolation. Possibly chosen with a prefix, e.g `r"foo\(\.\)+"`.

zuiderkwast: I like that Erlang becomes more Perl-like. :-)

wojtekmach: > Finally, choosing backtick will also depend on the documentation format. If we choose Markdown, the backticks in Markdown are used for code snippets. Which means that, if I want to write ``string:to_upper(`hello`)``

This will be also problematic with heredocs and fenced code blocks:

image

(I'm pasting an imagine because I don't know how to write it in Markdown so it renders correctly. :))

I'm personally a fan of code blocks done with 4 spaces indent but fenced blocks are sometimes useful too.

zuiderkwast: @wojtekmach Markdown allows any number of backticks, so if you need 3 backticks, just write it within 4 backticks. Perhaps Erlang could allow that trick too:

`````
-doc ````
Some function.

Example:

```
> 1 + 2
3
```
````
`````

(Note that the above example was enclosed in 5 backticks.)

Comment (williamthome) [reply eeps/eep-0064.md:33]: I like the backtick idea, but I'll prefer it to be used like in Javascript, where it permits to do interpolation, for example:
```
Bar = `bar`,
<<"foo bar">> = `foo ${Bar}`.
```
or 
````
-define(BAR, `bar`).

-doc ```
foo ${?BAR}
```
````
Or (why not) single backtick as an alternative to triple quotes using the same idea, the last backtick defines the indentation
```
-doc `
backtick 
multiline
unicode
binary
`

% Single line also a valid syntax
-doc `backtick unicode binary`
```

BTW, triple quotes give more control of the encoding (I don't see any problem writing the below)

```
<<"""
foo bar
"""/utf8>>
```

File: eeps/eep-0064.md [hunk 4]
```diff
@@ -0,0 +1,187 @@
...
+ignore the meaning of `\` if inside a code block.
+
+*Code Example*
+
+    >> X = """
+           This text \
+           has no indentation
+           """
+    
+    Equivalent to: <<"This text has no indentation\n">>
```
williamthome, line 155: Makes sense the possibility to comment after `\`?
For example:
```
"""
foo \ % Maybe I want to break this line to add a comment about it
bar
"""
```

williamthome: > I think that's ambiguous because `\` is no longer at the end. 

Oh, this is true and makes things tricky.

> Should `\%` be allowed then at the end? 

Oh, using `\%` makes sense to me.

> And, if so, how do you actually know `\` is at the end?

I think this is easy, I've implemented this now based on your idea of the `\%` on [this branch](https://github.com/williamthome/heredocs/tree/feat/comment) of a side project called `heredocs`. That's what the implementation outputs:
```
"foo bar" = """
            foo \% This is a comment and must be ignored
            bar\
            """
```

I don't know if this is a good idea, but it's possible.

josevalim: My concern is that the last example is still ambiguous with escaping the `%` character itself. :)

RaimoNiskanen: I do not understand why you want a comment in a multi line string.
All in it should be string content, otherwise it isn't very useful, right?

We want it for Markdown documentation, and Markdown uses trailing `\` for forcing a line break, which clashes head on with using trailing `\` for newline escape.

So I do not know if we want newline escape.  Or any escape codes at all, for that matter, since Markdown uses `\` for escaping things, just as the regular Erlang string syntax, and these may also collide...

josevalim: > We want it for Markdown documentation, and Markdown uses trailing \ for forcing a line break, which clashes head on with using trailing \ for newline escape.

`\` will escape the new line character. To force the line break, you need `\\` (because Markdown needs to see the `\` itself). So to me it makes logical sense: if you use `\\`, you are escaping the character that would escape/hide the line break. :)

zuiderkwast: Comment inside the string makes the syntax too complex IMO. It's already possible to put two strings side-by-side to concatenate them, e.g.

```Erlang
"foo " % this comment is ignored
"bar" = """
        foo \
        """
        %% this comment would be ignored too
        """
        bar\
        """.
```

williamthome: > We want it for Markdown documentation

Sorry @RaimoNiskanen, I commented without thinking in Markdown.
BTW, we can write comments in Markdown in [many different ways](https://stackoverflow.com/questions/4823468/comments-in-markdown).

michalmuskala:
> I'm not a fan of triple quoting for this. Providing an end marker avoids having to worry about escaping and whatnot. Something like this in a terminal:
> 
> ```
> <<EOF
> hello
> EOF
> ```

This would make Erlang syntax context-dependent. You wouldn't be able to parse this anymore with `yecc` as it is today, since you'd need to track the "end" quotes.

---

In general with triple quote syntax, I love it from Elixir. However, having used Rust quite extensively that doesn't have it, regular multi-line strings can serve this use-case relatively decently. With a good-enough integration with a formatter, it's not particularly annoying either. For example see how it's used in Rust Analyzer for testing - https://github.com/rust-lang/rust-analyzer/blob/f8dec25bd70cc3568069daf8c3d5f2a65e3aa4cb/crates/ide/src/move_item.rs#L193.

I'd say the much more useful feature is "raw strings" that disable escape sequences.

Taking examples from the EEP, they would need to be written as below:
```erlang
-doc "
Removes double-quotes (\") from a given string
".

foo() ->
    X = "
This is the beginning of the triple-quote text
",
    use_x(X).
```

The un-nesting of string under `X` is fairly awkward, but is significantly less painful than I initially anticipated seeing it for the first time.

This does leave the issue of the extra whitespace at the beginning (and end) though it could be handled with "escaping the newline":
```erlang
foo() ->
    X = "\
This is the beginning of the triple-quote text\
",
    use_x(X).
```

While this is "uglier" it does avoid introducing an new concept and syntactic form to the language.

josevalim:
@michalmuskala I haven't felt the need for triple quoted strings in Erlang so far *but* if the goal is to introduce docstrings then I think the need for them is well justified. Imagine writing examples in the documentation and having to escape every string in the documentation itself. So I think the `#"..."#` delimiters in the Rust example is pretty important too.

williamthome:
@michalmuskala My concern is to generalize this syntax because of the list of integers. How to deal with this:
```bash
1> [92,65,66,67].
"\\ABC"
```
If I understand you correctly, the compiler would automatically strip the 92.

zuiderkwast:
If it's only for `-doc`, it's not required to add a new string literal. We can just have a multi-line doc attribute, something like this:

    -doc.
    Removes double-quotes (\") from a given string
    -end.

... which would need to be handled by the scanner and be equivalent to `-doc "Removes double-quotes (\") from a given string".`

michalmuskala:
> @michalmuskala I haven't felt the need for triple quoted strings in Erlang so far _but_ if the goal is to introduce docstrings then I think the need for them is well justified. Imagine writing examples in the documentation and having to escape every string in the documentation itself. So I think the `#"..."#` delimiters in the Rust example is pretty important too.

Yes, I fully agree that "raw" strings or strings that disable escaping are important - more than triple-quoted strings. Their utility would be beyond docs - e.g. regex

williamthome:
@zuiderkwast It's not only for documentation, it's also useful to write better multi-line strings in Erlang code. See the [EEP abstract and rationale](https://github.com/erlang/eep/pull/47/files#diff-83e09021758096a1f356ce9f68db1fd98b95f067ec40eb8baa899c67fea80d13R11-R26). It's the main reason, but not only for this propose.
See also [this example](https://github.com/elixir-lang/elixir/blob/0872ee5444f257744866a4d4c237057f2d97d601/lib/elixir/src/elixir_parser.yrl#L1237-L1249) from @josevalim.

eeps/eep-0064.md
KennethL: I think this discussion has diverged too far from "what problem do we want to solve" with triple-quoted binary strings.
For me the problem to solve first is "How to write the documentation (Markdown assumed) in the `-doc` attribute in a convenient way. This is a different context from the program code in the same module. The text written in the `-doc` attribute is not to be available for the "program" , it should just land in a "doc-chunk" .

Another separate problem we want to solve is to have a nicer syntax for utf8 encoded binaries from <<"abc">> to something like `b"abc"`.

There are other things we want to improve as well and there might be an advantage to combine them, reuse notation etc. But first I think we should solve how to write documentation. What is written inside the `-doc` attribute is not part of the program in my view it is like a comment with the special property "part of the documentation" and its content does not matter for the ordinary compiler.

erszcz: Maybe I'm missing something, but I don't see triple quoted strings as a necessary prerequisite for [EEP-59](https://www.erlang.org/eeps/eep-0059). Wouldn't the following suffice?

````erlang
-module(z).

-doc "
This is an example module doc.

With embedded preformatted text:

    f() ->
        ok.

And also backtick-delimited text, too:

```erlang
g() ->
    ok.
```
".

f() -> ok.
````

All the formatting is preserved, so it's up to the Markdown engine to interpret it:

````
20> z:module_info(attributes).
[{vsn,[105165884523284496141303209755187911974]},
 {doc,"\nThis is an example module doc.\n\nWith embedded preformatted text:\n\n    f() ->\n        ok.\n\nAnd also backtick-delimited text, too:\n\n```erlang\ng() ->\n    ok.\n```\n"}]
21> io:format("~ts\n", [string:trim(ModuleDoc)]).
This is an example module doc.

With embedded preformatted text:

    f() ->
        ok.

And also backtick-delimited text, too:

```erlang
g() ->
    ok.
```
ok
22>
````

Indeed, stripping leading whitespace could allow for more flexibility when formatting the documentation in source code (so triple-quoted strings could come in handy if/when they arrive), but it doesn't seem to be a must have for EEP-59.

Comment (josevalim) [reply eeps/eep-0064.md:175]: Here is the issue in my opinion. Imagine writing docs for `string:uppercase/1`:

```erlang
-doc "
Converts `String` to uppercase.

See also `titlecase/1`.

## Examples

    1> string:uppercase(\"Michał\").
    \"MICHAŁ\".

".
```

Basically all of the `string` module would be quite annoying to write examples for. But I agree with @KennethL. The issue is documentation. Triple-quoted strings is one of the possible solutions.

williamthome:
Documentation is the main reason for this EEP and Markdown was mentioned a lot of times to be compatible and to not clash with it, but reading about Markdown to technical writings I ended to [AsciiDoc](https://docs.asciidoctor.org/asciidoc/latest):

> [...] AsciiDoc is a lightweight and semantic markup language primarily designed for writing technical documentation.
[...] An AsciiDoc processor can parse and comprehend an AsciiDoc source document and convert the parsed document structure into one or more output formats, such as HTML, PDF, EPUB3, man(ual) page, or DocBook. The ability to produce multiple output formats is one of the main advantages of AsciiDoc. 
[...] AsciiDoc bridges the gap between ease of writing and the rigorous requirements of technical authoring and publishing. AsciiDoc only requires a text editor to read or write, thereby offering a low bar to getting started.

The license is Apache 2.0. It's simple and powerful. O'Reilly has it in their [docs](https://docs.atlas.oreilly.com/writing_in_asciidoc.html) and has a repo with [examples](https://github.com/oreillymedia/orm_book_samples/tree/master/asciidoc_only) of how to, so books are written using it. Looking at the examples the format is compatible with GitHub.

I'm not trying to cause any confusion about this, quite the opposite. Maybe it can be a better alternative to Markdown for docs.
I saw @essen using AsciiDoc in [Cowboy](https://github.com/ninenines/cowboy/tree/master/doc/src/guide), so it's not something new for all.

[This link](https://docs.asciidoctor.org/asciidoc/latest/asciidoc-vs-markdown/) compares it to Markdown and [this link](https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#markdown-compatibility) talks about Markdown compatibility.

erszcz:
@williamthome For the context, please note a discussion on the choice of a documentation format has happened here: https://github.com/erlef/build-and-packaging-wg/issues/60

williamthome:
Good to know @erszcz, I hadn't seen this discussion. Thanks!

RaimoNiskanen:
Hi!

I rewrote the EEP to clarify the proposal.  I think a new read-through is needed.

https://github.com/erlang/eep/blob/21d4fc8cf61929260fab4a85f0ebf2dea434f4b4/eeps/eep-0064.md

Feedback is, again, welcome.

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+A later step, preferably done by the parser, uses the whitespace
+on the ending line as the definition of the string's indentation
+and strips that whitespace seqence from every line in the string,
+and strips the newline preceding the ending line.
+
+If any of the lines should not start with the defined indentation
+either because the line is too short or if the prefix differs,
+the parser reports a syntax error.
+
+Requiring that all lines must have exactly the same whitspace
```
erszcz, line 161:
```suggestion
Requiring that all lines must have exactly the same whitespace
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+  syntax errors.  Only a few will have a subtly changed
+  behaviour (string content).
+* Users can simply grep for `"""` in their source code.
+  Causing the same sequence e.g through macros would
+  be harder to find, but the worst problem would not
+  be new syntax errors (hard to miss), but changed
+  behaviour.  And the changed behaviour would be
+  a slightly different string content.
+
+Therefore, it should be very unlikly that anyone
```
erszcz, line 284:
```suggestion
Therefore, it should be very unlikely that anyone
```

RaimoNiskanen:
I ran it through a spell checker...

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+    Post-History:
+****
+EEP 64: Multi-line Indented Verbatim Strings
+----
+
+Abstract
+========
+
+This EEP proposes the introduction of Multi-line Indented Verbatim strings,
+*MIV Strings* and defines their semantics.  The main benefit is to allow
```
paulo-ferraz-oliveira, line 16:
```suggestion
*MIV Strings*, and defines their semantics.  The main benefit is to allow
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+*multi-line* strings in an easy and useful, *indented*, way,
```
paulo-ferraz-oliveira, line 17:
```suggestion
*multi-line* strings in an easy, useful, and *indented* way,
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+similar to in other languages, e.g Elixir.
```
paulo-ferraz-oliveira, line 18:
```suggestion
similar to other languages, e.g. Elixir.
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+
+Their first use case is for in-module documentation attributes
+containing [Markdown][] or similarly formatted text where *verbatim* text
```
paulo-ferraz-oliveira, line 21: Is `[Markdown][]` expected? It doesn't seem to be valid Markdown.

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+    Second line with \"\\*not emphasized\\* Markdown\"
+    Third line",
+                 {ok, X}
+        end.
+
+The content's indentation cannot adhere to the surrounding code's
+and the `*` has to be doubly escaped to get a  `\*` character
+sequence into the actual content.
+
+In a documentation attribute as suggested in [EEP 59][]
```
paulo-ferraz-oliveira, line 44: Is `[EEP 59][]` as expected?

paulo-ferraz-oliveira, line 44:
```suggestion
In a documentation attribute, as suggested in [EEP 59][],
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+
+    -doc """
+        First line
+        Second line with "\*not emphasized\* Markdown"
+        Third line
+        """.
+
+The main reason to consider this EEP is for documentation
+attributes, and here not having to worry about escape sequences
+is the most attractive property of this EEP.  But introducing
```
paulo-ferraz-oliveira, line 56:
```suggestion
attributes, an not having to worry about escape sequences
is its most attractive property. But introducing
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+a new string format will also require defining how it would behave
+in Erlang code.
+
+Having a string format only allowed in attributes would simply be
```
paulo-ferraz-oliveira, line 60:
```suggestion
Having a string format only be allowed in attributes will simply be
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+    -doc """
+          Better formatted
+        documentation paragraph
+        /-\
+        \-/
+        """.
+
+### MIV String Scanner Token
+
+A MIV String must be a token that the scanner recognizes as a string,
```
paulo-ferraz-oliveira, line 90:
```suggestion
An MIV String must be a token that the scanner recognizes as a string,
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+which makes it suitable for a documentation attribute value term.
+It starts and ends with three double quotes: `"""`.
+
+Double quotes, `"`, are chosen because normal Erlang strings use them
+and this is just a new variant.  Since double quotes are used
+a MIV String should, as a normal string, produce a list of codepoints.
```
paulo-ferraz-oliveira, line 96:
```suggestion
and this is just a new variant.  Since double quotes are used,
an MIV String should, as a normal string, produce a list of codepoints.
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+that consists of a `-` token, an atom, one value term and a full stop
+(dot).  The value term may be enclosed in parentheses
+(which is not very interesting for documentation attributes).
+
+    -doc "  Badly formatted
+    documentation paragraph
+    /-\\
+    \\-/".
+
+A documentation attribute should have a string as its content term,
```
paulo-ferraz-oliveira, line 77:
```suggestion
A documentation attribute shall have a string as its content term,
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+or we can use the GitHub [Markdown][] trick to allow
+3 or more start characters and matching end characters
+so this would be valid:
+
+    X = """"
+        A MIV String starts with:
+        """
+        and ends with:
+        """
+        """"
```
williamthome, line 327: I liked the four double quotes idea.
I've updated [my PR](https://github.com/erlang/otp/pull/7313) with this change. Now this is possible:
```bash
1> """"
.. ```erlang
.. foo() ->
..     """
..     foo
..     bar
..     """.
.. ```
.. """".
"```erlang\nfoo() ->\n    \"\"\"\n    foo\n    bar\n    \"\"\".\n```\n"
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+It starts and ends with three double quotes: `"""`.
+
+Double quotes, `"`, are chosen because normal Erlang strings use them
+and this is just a new variant.  Since double quotes are used
+a MIV String should, as a normal string, produce a list of codepoints.
+
+It would be more convenient if a MIV String produced an UTF-8 binary,
+but that would be surprising when using double quotes, and
+the documentation build process can convert the codepoint list
+into the needed binary chunk.
```
paulo-ferraz-oliveira, line 101:
```suggestion
It will be more convenient if an MIV String produces an UTF-8 binary,
but that will be surprising when using double quotes, and
the documentation build process can convert the codepoint list
into the needed binary chunk.
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+
+In source code a MIV String is valid in a binary so
```
paulo-ferraz-oliveira, line 103:
```suggestion
In source code, an MIV String is valid in a binary, so
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+
+A MIV String ends with white-space on a new line followed by `"""`.
+This completes the scanner token.
+
+A later step, preferably done by the parser, uses the white-space
+on the ending line as the definition of the string's indentation
+and strips that white-space sequence from every line in the string,
+and strips the newline preceding the ending line.
+
+If any of the lines should not start with the defined indentation
```
paulo-ferraz-oliveira, line 157:
```suggestion
If any of the lines do not start with the defined indentation
```

eeps/eep-0064.md
wojtekmach: It is. Just `[Markdown]` works too but it might be a GitHub-Flavoured Markdown thing.

essen:
I suppose the `""""` can be a good middle ground but that's quickly going to become unreadable so perhaps the maximum number should be limited (as a convention maybe).

I have one snippet that conflicts with the way your MIVs are currently defined, and that is:

```
	#{code := 200, body := <<"true">>} = do_raw(Config, [
		"POST /echo/has_body HTTP/1.1\r\n"
		"Host: localhost\r\n"
		"Content-length: 12\r\n"
		"\r\n"
		"Hello world!"]),
```

I have these kinds of strings all over my tests and if I understand correctly they cannot be expressed because the \r is removed from the line ending and it's not possible to include it in the MIV string (no escaping). It's a bit of a shame, it would fit multi-line strings well, and [I got a lot of them](https://github.com/ninenines/cowboy/blob/master/test/rfc7230_SUITE.erl).

Not that this should be a blocker or anything, I think I prefer explicit line endings since these are tests, but food for thoughts.

eeps/eep-0064.md
paulo-ferraz-oliveira: [Markdown][] < `[Markdown][]`
[Markdown] < `[Markdown]`

^ doesn't seem to do anything in particular in GitHub-Flavoured Markdown.

wojtekmach: I don't know what the above means but please not that in order for `[ref]` and `[ref][]` to link, you need a `[ref]: url` somewhere in your Markdown.

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+
+If any of these characters is not white-space,
+the parser reports a syntax error.
+
+### MIV String End
+
+All bytes are collected as they are (verbatim) and becomes
+the MIV String content.
+
+A MIV String ends with white-space on a new line followed by `"""`.
```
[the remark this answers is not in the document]
josevalim:
```suggestion
An MIV String ends with a new line followed by optional white-space and then by `"""`.
```

eeps/eep-0064.md
paulo-ferraz-oliveira: > I don't know what the above means

It was to show you (to the left) how stuff was rendered, but José's reply enlightened me. I've used this pattern before, mostly in `README.md`s, but somehow this just fell odd seeing it today 😄 Must be Friday-tiredness...

williamthome:
The trailing backslash was dropped in the last change.
IMHO it's an important feature to keep the code organized and readable and we can use it without clashing with the Markdown.
I believe the reason is that the backslash is used in Markdown to break a line, the same as the `<br>` tag in HTML. Double whitespace at the end of the line also has the same behavior, but it's not recommended.
Some notes from [Google Markdown Style Guide](https://google.github.io/styleguide/docguide/style.html#trailing-whitespace) about this:
> Don’t use trailing whitespace, use a trailing backslash.
>
> The [CommonMark spec](http://spec.commonmark.org/0.20/#hard-line-breaks) decrees that two spaces at the end of a line should insert a <br /> tag. However, many directories have a trailing whitespace presubmit check in place, and many IDEs will clean it up anyway.
>
> Best practice is to avoid the need for a <br /> altogether. Markdown creates paragraph tags for you simply with newlines: get used to that.

The recommendation is to use the return to create a newline as also [described here](https://www.markdownguide.org/basic-syntax/#line-break-best-practices):
> You can use two or more spaces (commonly referred to as “trailing whitespace”) for line breaks in nearly every Markdown application, but it’s controversial. It’s hard to see trailing whitespace in an editor, and many people accidentally or intentionally put two spaces after every sentence. For this reason, you may want to use something other than trailing whitespace for line breaks. If your Markdown application [supports HTML](https://www.markdownguide.org/basic-syntax/#html), you can use the <br> HTML tag.
>
> For compatibility, use trailing white space or the <br> HTML tag at the end of the line.
>
> There are two other options I don’t recommend using. CommonMark and a few other lightweight markup languages let you type a backslash (\) at the end of the line, but not all Markdown applications support this, so it isn’t a great option from a compatibility perspective. And at least a couple lightweight markup languages don’t require anything at the end of the line — just type return and they’ll create a line break.

I'll give some points of view about it using [this chunk of text](https://github.com/erlang/eep/pull/47/files#diff-83e09021758096a1f356ce9f68db1fd98b95f067ec40eb8baa899c67fea80d13R103-R117) from this EEP, an [online Markdown editor](https://pandao.github.io/editor.md/en.html) and the Erlang shell on the branch of the [PR that I submitted for this EEP](https://github.com/erlang/otp/pull/7313).

### Just copying the text

Ok, this obviously will not produce an expected result, but imagine you want to limit to 72 the length of chars in your editor and you want to type the text that I mentioned previously. Just copying the text will more or less simulate a max column length mark of an editor.

```
1> io:format("~ts~n", [""""
.. In source code a MIV String is valid in a binary so
.. producing a Unicode binary is reasonably straightforward:
.. 
..     X = <<"""
..         Line 1
..         Line 2
..         """/utf8>>
.. 
.. The extra overhead is not exhausting since this we are targeting
.. multi-line strings here.
.. 
.. As a future expansion it has been proposed to use prefixes
.. for specialized strings such as regular expressions,
.. interpolated variables ([PR-7343][]), Unicode binary strings, etc.
.. For example: `X = u"Tschüß"` for an UTF-8 encoded string.
.. """"]).
In source code a MIV String is valid in a binary so
producing a Unicode binary is reasonably straightforward:

    X = <<"""
        Line 1
        Line 2
        """/utf8>>

The extra overhead is not exhausting since this we are targeting
multi-line strings here.

As a future expansion it has been proposed to use prefixes
for specialized strings such as regular expressions,
interpolated variables ([PR-7343][]), Unicode binary strings, etc.
For example: `X = u"Tschüß"` for an UTF-8 encoded string.

ok
```

image

### Rewriting to an expected result

Removing the unexpected newlines we'll need to scroll to see the text. I imagine no one like to code scrolling horizontally.

```
2> io:format("~ts~n", [""""
.. In source code a MIV String is valid in a binary so producing a Unicode binary is reasonably straightforward:
.. 
..     X = <<"""
..         Line 1
..         Line 2
..         """/utf8>>
.. 
.. The extra overhead is not exhausting since this we are targeting multi-line strings here.
.. 
.. As a future expansion it has been proposed to use prefixes for specialized strings such as regular expressions, interpolated variables ([PR-7343][]), Unicode binary strings, etc.
.. For example: `X = u"Tschüß"` for an UTF-8 encoded string.
.. """"]).
In source code a MIV String is valid in a binary so producing a Unicode binary is reasonably straightforward:

    X = <<"""
        Line 1
        Line 2
        """/utf8>>

The extra overhead is not exhausting since this we are targeting multi-line strings here.

As a future expansion it has been proposed to use prefixes for specialized strings such as regular expressions, interpolated variables ([PR-7343][]), Unicode binary strings, etc.
For example: `X = u"Tschüß"` for an UTF-8 encoded string.

ok
```

image

### Using the trailing backslash

I see the code below much more readable and produces the same result as the previous text. Also, no extra line is produced at the end.
> **Note**
> The scroll is caused by the resulting text.

```
3> io:format("~ts~n", [""""
.. In source code a MIV String is valid in a binary so \
.. producing a Unicode binary is reasonably straightforward:
.. 
..     X = <<"""
..         Line 1
..         Line 2
..         """/utf8>>
.. 
.. The extra overhead is not exhausting since this we are targeting \
.. multi-line strings here.
.. 
.. As a future expansion it has been proposed to use prefixes \
.. for specialized strings such as regular expressions, \
.. interpolated variables ([PR-7343][]), Unicode binary strings, etc.
.. For example: `X = u"Tschüß"` for an UTF-8 encoded string.\
.. """"]).
In source code a MIV String is valid in a binary so producing a Unicode binary is reasonably straightforward:

    X = <<"""
        Line 1
        Line 2
        """/utf8>>

The extra overhead is not exhausting since this we are targeting multi-line strings here.

As a future expansion it has been proposed to use prefixes for specialized strings such as regular expressions, interpolated variables ([PR-7343][]), Unicode binary strings, etc.
For example: `X = u"Tschüß"` for an UTF-8 encoded string.
ok
```

image

### To completeness

The backslash in Markdown can be used normally escaping it. For example:

```
4> io:format("~ts~n", ["""
.. This line ends here.\\
.. This is a new line.\
.. """]).
This line ends here.\
This is a new line.
ok
```

Sorry for this long comment and please correct me if I've missed something :)

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+
+The main reason to consider this EEP is for documentation
+attributes, and here not having to worry about escape sequences
+is the most attractive property of this EEP.  But introducing
+a new string format will also require defining how it would behave
+in Erlang code.
+
+Having a string format only allowed in attributes would simply be
+very strange and the one suggested in this EEP would be useful
+also in Erlang code.
```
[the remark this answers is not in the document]
zuiderkwast: I believe "would" is the right word here. It's conditional, with an implicit if, such as "if it would be included in Erlang".

eeps/eep-0064.md
zuiderkwast: C# has this three-or-more double quotes syntax too.

eeps/eep-0064.md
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I saw it as "being implemented" thus my choice, but I agree that if it's not for "now", "would" is probably the better choice. I'll update it.

eeps/eep-0064.md
RaimoNiskanen: I'll go for `easy and useful (*indented*) way,`, which is more according to my intention

eeps/eep-0064.md
```diff
@@ -0,0 +1,354 @@
...
+    -doc """
+        First line
+        Second line with "\*not emphasized\* Markdown"
+        Third line
+        """.
+
+The main reason to consider this EEP is for documentation
+attributes, and here not having to worry about escape sequences
+is the most attractive property of this EEP.  But introducing
+a new string format will also require defining how it would behave
```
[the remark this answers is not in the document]
RaimoNiskanen: Rephrased more or less like you suggested to:
```
attributes, where not having to worry about escape sequences
is this EEP's most attractive property.  Introducing a new string
format, however, will also require defining how it shall behave
in Erlang code.
```

eeps/eep-0064.md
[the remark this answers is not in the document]
RaimoNiskanen:
```
Having a string format that is only allowed in attributes would
simply be very strange and the one suggested in this EEP
would also be useful in Erlang code.
```

eeps/eep-0064.md
RaimoNiskanen: Since this depends on if the correct pronunciation of the abbreviation "MIV" is "emm eye vee", "miv" ("meev") or "multiline ...", which is unclear given that it is not an established abbreviation, I think I will be creative and rename the whole shebang to "Verbatim Multi-line Indented Strings", also because the EEP argues that "Verbatim" is the most urgent property.

eeps/eep-0064.md
RaimoNiskanen: @williamthome: Note that the idea is to allow 3 or more, not 3 or 4...

eeps/eep-0064.md
RaimoNiskanen: This is conditional, and I argue *against* producing a binary, so it *would* be more convenient, but it is a bad idea for other reasons.

I will rephrase a bit, though.

RaimoNiskanen:
@essen 
> I have these kinds of strings all over my tests and if I understand correctly they cannot be expressed because the \r is removed from the line ending and it's not possible to include it in the MIV string (no escaping).

Yes or no.  It should be possible to use Verbatims strings for these test strings.

Your editor should allow you to insert CR into the source code, last on the line.  In Emacs you use Ctrl-Q to quote the CR ([Enter]).  It is only on the last line; where `CR? LF Whitespace* " " "` is considered to be the string ending.  So CR LF is stripped only from the last content line.

All whitespace (`CR` `LF` `TAB` `HT` ...) on the preceding lines is copied verbatim.

Also, the string start is `" " " Whitespace* LF`, which is not part of the string content.

Note that `CR` is considered `Whitespace` in the scanner, but `LF` is end of line.

RaimoNiskanen:
@williamthome: Regarding trailing backslash and escaping.  Just because of all the quirks of escape sequences e.g. by backslashes in different documentation text formats such as Markdown, maybe also AsciiDoc, and whatnot, I think it is simplest to not have any escape character in these Verbatim strings.

We'll see how to do about value interpolation one day...

RaimoNiskanen:
Thank you for the feedback!  I have updated (and renamed, again).

I realized that there are two forms of empty string, so I wrote a paragraph about that.
My suggestion is to allow both forms.

essen:
> @essen
> 
> > I have these kinds of strings all over my tests and if I understand correctly they cannot be expressed because the \r is removed from the line ending and it's not possible to include it in the MIV string (no escaping).
> 
> Yes or no. It should be possible to use Verbatims strings for these test strings.

Alright then it is possible but it is not ideal: the CR is hidden (like the LF), so it's no longer possible to see the difference between a test that tries sending requests with only LF, compared to a test that properly sends CRLF for example. You can see it if you display hidden whitespaces but you won't see it on GitHub or others.

Modifying the string to make the LF into CRLF would make things more obvious I think (`crlf(""" ... """)`). Perhaps this warrants an example in the EEP.

RaimoNiskanen:
@essen 
> You can see it if you display hidden whitespaces but you won't see it on GitHub or others.
> 
That is a problem.  In Emacs it shows as `^M`, unless, I guess in *DOS* mode where it probably is hidden.  So Emacs is an exception, and that is not enough.

> Modifying the string to make the LF into CRLF would make things more obvious I think (`crlf(""" ... """)`). Perhaps this warrants an example in the EEP.

There are many suggestions on creative string prefixes appearing.

`b"` `u"` `re"` `crlf"` `f"` `d"`, ...
Binary, Unicode, Regular Expression, CR-LF-expanded, Formatted (Interpolated), Debug Formatted (Verbose Interpolation), and probably more.

How to combine them into a reasonable semantics will have to be a future problem.  This EEP mentions "specialized strings such as regular expressions, interpolated variables ([PR-7343][]), Unicode binary strings, etc" and I think this can fall under other "specialized strings".

Your use case needs escape sequences, so as you do concatenating single quoted strings is maybe not that bad...

essen:
@RaimoNiskanen 

> `crlf"`

Probably not a good idea anyway. CRLF is increasingly becoming a legacy problem. I wouldn't build a feature around it. Workarounds in the documentation are appreciated though.

> Your use case needs escape sequences, so as you do concatenating single quoted strings is maybe not that bad...

The downside is not being able to just copy paste stuff, for example examples from RFCs or output from developer tools. Or copying a test input into `netcat` to try it in a terminal. So I think the `crlf(""" ... """)` solution would be the best fit for this particular scenario (tests).

eeps/eep-0064.md
```diff
@@ -6,16 +6,16 @@
...
 
 Abstract
 ========
 
-This EEP proposes the introduction of Multi-line Indented Verbatim strings,
-*MIV Strings* and defines their semantics.  The main benefit is to allow
-*multi-line* strings in an easy and useful, *indented*, way,
-similar to in other languages, e.g Elixir.
+This EEP proposes the introduction of Verbatim Multi-line Indented strings,
+*VMI Strings*, and defines their semantics.  The main benefit is to allow
```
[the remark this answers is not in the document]
RaimoNiskanen: @essen: I guess the fitting general feature here would be Multi-line Indented string, that has escape sequences.  The question is if that merits a prefix...

@zuiderkwast: I just wanted to call them something in the EEP that was not too long since I have to mention them many times.  "Triple quoted strings" is a better name, even if they could be "N>=3 quoted strings".  "Verbatim strings" is another possibility.  But if we later add prefixes to make them more flexible, e.g. loose "verbatim",  it is maybe "triple quoted" that is the one property that must remain...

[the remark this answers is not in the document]
RaimoNiskanen: "Triple quoted" or "Multi-line".  I'd say we cannot take that away from them either. That they can be indented is just what makes them good for multi-line.

RaimoNiskanen:
I have polished two small language details.  That, I hope, would be the last on this track.

Now, in the next commit, I will switch viewpoint and argue for multi-line strings with escape sequences...

eeps/eep-0064.md
williamthome: @RaimoNiskanen Oh, nice! I'll change the code to support 3+ double quotes.

williamthome: The PR is updated and now 3+ double quotes are supported.

RaimoNiskanen:
> Now, in the next commit, I will switch viewpoint and argue for multi-line strings with escape sequences...

That did not happen.  I renamed them, again, to "triple-quoted strings" and clarified some details.

I tried to see that escape sequences would be useful, but ended up back in that it will be nice to have them out of the way for any documentation text markup source format

williamthome:
@RaimoNiskanen Now I see what you mean and I agree with you. This EEP should not include the trailing backslash or any other, this is a job for a post-process or format functions. Verbatim strings should only handle multiline strings and escape quotes.
An example where escaped sequences can be a problem is if the `-doc` attribute be implemented and a file as argument be accepted, for example
```erlang
-doc {file, "foo.md"}.
```
and the `foo.md` content be 
```markdown
This is the first line.\
This is the second one.
```
considering the backslash trailing
```erlang
% this will be the output:
-doc "This is the first line.This is the second one."

% but this is expected
-doc "This is the first line.\\\nThis is the second one."
```
So, I switch viewpoint and I'm against the trailing backslash that I mentioned before.

zuiderkwast:
@williamthome quotes cannot be escaped. Nothing can be escaped.

essen:
> I tried to see that escape sequences would be useful, but ended up back in that it will be nice to have them out of the way for any documentation text markup source format

Yes I'm arriving at the same conclusion. It will be very useful to not have to escape anything, and not just for documentation. Strings can always be post-processed if necessary (even as a parse transform if that's important).

eeps/eep-0064.md
```diff
@@ -0,0 +1,395 @@
...
+
+This means that the examples in the following text
+where a normal string is used as matching reference
+assumes that the source code has `LF`-only newlines,
+such as:
+
+    "\nX" = """
+        
+        X
+        """"
```
williamthome, line 196:
```suggestion
        """
```

eeps/eep-0064.md
```diff
@@ -0,0 +1,395 @@
...
+
+If the source code has `CR` `LF` newlines that
+example instead becomes:
+
+    "\r\nX" = """
+        
+        X
+        """"
```
williamthome, line 204:
```suggestion
        """
```

williamthome (changes requested):

eeps/eep-0064.md
```diff
@@ -0,0 +1,395 @@
...
+    This is a syntax error (incorrect indentation)
+        """,
+    
+    """ This is a syntax error
+    (non-white-space on start line)
+    """,
+    
+    """
+    This will probably be a syntax error
+    since the ending line is not here"""
```
williamthome, line 269:
```
    """
    This will probably be a syntax error
    since the ending line is not here"""
```
IMO this is not totally correct. It's a syntax error, but the shell would wait for the string ending, and in a module the parser would give up, for example:
```erlang
-module(foo).
-export([foo/0]).

foo() ->
    """
    This will cause the parser to give up
    since the ending line is not here""".
```

```suggestion
    """
    This will cause the shell to wait for the `"""` in a new line or 
    the parser to give up in a module, since the ending line is not here"""
```

williamthome (changes requested):

eeps/eep-0064.md
RaimoNiskanen: Oops!

eeps/eep-0064.md
RaimoNiskanen: I will rephrase that

RaimoNiskanen:
I added the sections "Verbatim strings" and "Comparison with Elixir"

josevalim:
For what is worth, I am very happy with the current proposal. It outlines the problem well, the design decisions and their limitations, plus possible future improvements. :+1: Regarding the name, I would propose indeed verbatim strings, as it clarifies it does not support escape characters.

RaimoNiskanen:
I also think "Verbatim Strings" is a better name, today, but given that a future extension might be Sigils, these Triple-quoted Strings could start to support e.g. escape sequences and interpolation...

So "triple-quoted" is the thing that they always will be.  Except maybe if they become "at-least-triple-quoted", but the name is still sort of valid.

"Heredocs" could be a name, especially for a free-to-choose delimiter as at-least-triple-quoted.  I actually think the name "heredocs" in Elixir is a bit misleading since the delimiter is fixed to `"""`, but in shells, where the name originates, the delimiter is a free-to-choose word.

eeps/eep-0064.md
```diff
@@ -1,176 +1,480 @@
...
+
+In source code a triple-quoted string is valid within a binary,
+so producing a Unicode binary is reasonably straightforward:
+
+    X = <<"""
+        Line 1
+        Line 2
+        """/utf8>>
+
+The 2 + 7 characters overhead is not exhausting since we are targeting
```
kikofernandez, line 173: I know this is obvious to most people but I do not follow where these characters 2 + 7 mean or come from.
Not sure if you can explain offline or here, so that the EEP has all content

eeps/eep-0064.md
```diff
@@ -0,0 +1,491 @@
...
+
+### Triple-Quoted String End
+
+All characters are collected as they are (verbatim) and becomes
+the string content.
+
+A triple-quoted string ends with newline followed by optional white-space
+and then `"""`.  This completes the scanner token.
+
+A later step, preferably done by the parser, uses the white-space
```
zuiderkwast, line 214: Is it really preferable to strip indentation in the parser? I can't see why this shouldn't be done in the lexer. If the lexer strips indentation, the token appears with the actual string content in the sequence of tokens, as with normal strings where `erl_scan` handles the escape sequences.

eeps/eep-0064.md
essen: The 2 is `<<` and the 7 is `/utf8>>` which precede/follow the triple quotes.

eeps/eep-0064.md
RaimoNiskanen: That would be convenient, we maybe wouldn't even need a new token, but having the lexer/scanner strip indentation would mean an extra pass over the string's content, which is something the scanner doesn't do today, so it would complicate the scanner structure.

Furthermore, I suspect that to have a source code preserving compiler (robust parser), which is on our ToDo list, also skipped characters should be passed through the parser to the abstract format.

But there might be no actual strong reasons against doing it in the scanner.  We can start there and see what happens...

RaimoNiskanen:
I merged this EEP which only means that it is recognized and in status: Draft

eeps/eep-0064.md
josevalim: I am a bit late to the party but I like the idea a lot too. This and automatically stripping the last line are elegant solutions to the problems presented. :+1:

PR #49, opened 2023-06-29
Updates of EEP 64

RaimoNiskanen:
3-or-more-quoted strings may have potential for
bigger backwards compatibility.


RaimoNiskanen:
The reference implementation of this EEP is in erlang/otp#7451

PR #53, opened 2023-09-26
Add EEP 66: Sigils for String Literals

RaimoNiskanen:


eeps/eep-0065.md
```diff
@@ -0,0 +1,282 @@
...
+
+[PR-7684][] Implements the basics of handling Sigils on string literals.
+The scanner produces a «`sigil`» token before the string literal, and a
+«`sigil_suffix`» token after.  These are then removed by the parser.
+
+Another approach would be to produce e.g a «`sigil_string`» token
+for the whole string and then handle that in the parser.
+It would require more state to be kept in the scanner between
+the parts of the sigil prefixed string, and therefore need
+more scanner rewriting.
```
josevalim, line 231: The reason why Elixir introduces lower-cased and upper-cased sigils is because you can import custom functions as sigils. `~FOO["bar"]` is equivalent to calling `sigil_FOO("bar")`. Therefore, because interpolation is a tokenizer/parser feature, the name needs to tell us if the feature is enabled or disabled.

On the Erlang side, you could implement this by emitting new AST nodes. This means someone can use a parse transform (or something else) to give meaning to new sigils.

However, if the feature is fully implemented on the parser/tokenizer and is not extensible, then you know upfront if a sigil supports interpolation or escape codes. This means you don't need to rely on the lower/upper case sigil rule. You could use `unescaped_string"foo"` or even introduce `s"foo"U` or some other letter to mean escaping is disabled (although having such important information at the end of the sigil is weird).

josevalim (commented):
Hi @RaimoNiskanen!

I have dropped some comments on the design decisions behind Elixir sigils. I am not sure how relevant they are but hopefully the context helps a bit.

In a nutshell, Elixir sigils can sometimes be too concise and carry information in hidden ways (such as using uppercase/lowercase letter to denote escaping/unescaping) and, while those choices were made for valid reasons in Elixir (at least I hope so!), Erlang may choose to make more explicit choices. :)

If you think there are other ways I can help, feel free to reach out! 💯

eeps/eep-0065.md
```diff
@@ -0,0 +1,282 @@
...
+regular expressions can make use of suffix flags to control
+expression compilation such as the «`i`» flag for a character case
+insensitive regular expression.
+
+### Sigil Type
+
+The suggested Sigil Types, inherited from Elixir, are:
+
+* «`s`»: binary string, an string represented as an UTF-8 encoded binary
+* «`c`»: char list, an Erlang string (Unicode codepoint list)
```
josevalim, line 163: Nitpick: the [types](https://www.erlang.org/doc/man/unicode#type-charlist) say `charlist` instead of `char_list`, so it may be worth making the spacing consistent here:

```suggestion
* «`c`»: charlist, an Erlang string (Unicode codepoint list)
```

eeps/eep-0065.md
```diff
@@ -0,0 +1,282 @@
...
+
+Here we define a Sigil as a prefix to a string literal
+that indicates how to interpret the string literal.
+It consists of:
+
+1. The character «`~`»
+2. An (optional) character in the range «`a`»-«`z`» or «`A`»-«`Z`»,
+   that specifies the [Sigil Type][].  The Sigil Type can
+   affect how the String Content is scanned, for example
+   if escape sequences are honoured.
```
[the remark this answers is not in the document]
RaimoNiskanen: Multi-letter Sigils in Elixir, indeed... Good to know!

In Erlang there should be no problem to allow multi-letter sigils without uppercase restriction.

Word lists, time formats and custom sigils are among the ones I think are the least interesting for Erlang.  The ones in the EEP I think are the interesting, and for them (binary, interpolation, and regexp) it would be attractive that the sigil is short:  `~utf8_binary"Content"` is not obviously better than `<<"Content"/utf8>>`, but `~s"Content"` is.

So I don't think those special Sigils are strong arguments to allow multi-letter Sigils.

That said; my first internal suggestion was that a Sigil String Literal would be `~xyz""` with no suffix, and even regexp flags should be handled in the prefix.  And I still see that it would be nice to have interpolation as a character flag in the prefix, making it easier to introduce Sigils before deciding on interpolation.

Still,  see no really good reason to limit the sigil prefix to one character - I only copied what I thought was Elixir's syntax.  But I find more reason to allow multi-letter sigils in order to to have more flag characters in the prefix, than to have a more descriptive name as the sigil prefix.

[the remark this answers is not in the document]
josevalim: > it would be attractive that the sigil is short: ~utf8_binary"Content" is not obviously better than <<"Content"/utf8>>, but ~s"Content" is.

I agree.

---

It also all depends on what is the extensibility story. If the sigils are all hardcoded in the tokenizer/parser, then you can keep it single-letter and make it to multi-letter when needed. But if you want to allow custom sigils via parse transforms, then allowing the community to be clearer with longer names is a huge plus. And I would say some of the best use cases of sigils in Elixir comes from the community:

1. Phoenix uses sigils for writing routes, such as: `~p"/posts/#{id}"` - It works the same as `~s"/posts/#{id}"` except that Phoenix also validates that the route exists, helping developers catch typos (such as if someone wrote ~p"/post/#{id}" instead).

2. Nx uses `~MAT` and `~VEC` for writing matrices and vectors. Instead of:

```elixir
tensor([
  [1.0, 0.0, 2.0],
  [3.0, 5.0, 4.0],
  [5.0, 4.0, 6.0]
])
```

you can write:

```elixir
~MAT"
  1.0 0.0 2.0
  3.0 5.0 4.0
  5.0 4.0 6.0
"
```

Besides helping with readability, it also converts the tensor into its underlying binary format at compile-time.
You can also specify the type+bit at the end: `~MAT[...]f16`

3. Phoenix uses `~H` (but it may have been `~HTML` if we could go back in time) for writing HTML templates. It supports interpolation and validates the contents is actually valid HTML. You can also see someone using this to write valid XML fragments, etc

eeps/eep-0065.md
```diff
@@ -0,0 +1,282 @@
...
+
+In a general sense, a [Sigil][3], is a prefix to a variable
+that indicates its type, such as «`$I`» in Basic or Perl,
+ where «`$`» is the sigil and «`I`» is the variable.
+
+Here we define a Sigil as a prefix to a string literal
+that indicates how to interpret the string literal.
+It consists of:
+
+1. The character «`~`»
```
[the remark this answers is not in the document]
RaimoNiskanen: The other reason I do not like these simple prefixes is because the implications they have on the structure of the language, in that how they affect the lexer and the parser.

I will write a section on my woes about the lexer and the parser, in the EEP.

Today, `bf"my regex"` is scanned (by the lexer) as atom:`bf`, string:`my regex`.  The lexer knows after scanning `b` that it is scanning an atom, and at `"` the atom ends since a new token, a string, starts.

With a simple prefix the lexer has to know which atom start sequences that may be string prefixes, or has to know that if an atom (or variable) ends with `"`, `'` (or `[`) it is a string prefix, not an atom (nor a variable).  This also introduces a small discrepancy between the first and the other characters in the prefix since `0`-`9` and `@` are not allowed as the first character.

I think it would be a really ugly odd bird if how a token ends changes the token type...

Can we leave this for the parser to figure out?

Then the parser would see that atom+string or variable+string should be a string prefix.  This opens for using a quoted_atom+string, which allows a much larger character set in the atom, and looks really odd: `'Bää' "Content"`.  The combination is a parse error today, and who knows which programmer mistakes we open for by allowing it.

But the biggest problem would be that the lexer did not know how to scan the string with respect to e.g escape sequences.  So either the lexer would have to peek at the atoms and variables it scans, and keep that in its state for a string that might follow, which would be doing the job of the parser, or the parser would have to translate escape sequences in the string, which is the job of the lexer.  ... The lexer would also have to know that after an atom or a variable, a quoted atom is not an atom but a string, and possibly a `[` is not a punctuation character, but instead the start of a string.  Again, the parser's job.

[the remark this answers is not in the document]
josevalim: > I think it would be a really ugly odd bird if how a token ends changes the token type...

I read this sentence and I thought: well, don't you already change the token based on the presence of the `@` character? And then I tried it out in the shell:

```erlang
1> foo@bar(bar).
** exception error: undefined shell command foo@bar/1
```

And of course `foo@bar` is a valid function name, which is simple and it makes total sense! In other words, I understand it better now, thank you.

> But the biggest problem would be that the lexer did not know how to scan the string with respect to e.g escape sequences. 

Just to close the loop on this conversation, depending if you want to allow custom sigils or to make them extensible, that's a decision you will have to make anyway. That's because the sigil is the one that decides the escape characters. After all, the `r` sigil uses different escape sequences than `s`.

So there is an option of saying that the sigil goes as a raw string all the way up to the Erlang Abstract Format, and the compiler/parse transform escapes and interpolates the sigil contents as they prefer. This would give you the maximum extensibility.

Elixir leaves the encoding up to the sigil (and [we provide helpers functions](https://hexdocs.pm/elixir/Macro.html#unescape_string/1)) but interpolation is made part of the tokenizer. Erlang is much simpler to tokenize/parse, so you could leave both of them up to the sigil author, if you want to.

eeps/eep-0065.md
RaimoNiskanen: I think we want to implement this feature in the lexer+parser and not have it extendable, other than through parse transforms.

It would be natural for the lexer to remember the sigil prefix when scanning the string, so it then knows how to handle escape codes (and interpolation).  If that information would be in the suffix, the scanner would have to first do a raw scan of the string, and then re-scan it if the suffix says that escape codes should be honoured.  That would be awkward and disturb error reporting since e.g a string end error would be detected before an escape sequence error that is located before the string end in the code.

So seems better if it is only the sigil prefix that affects how the lexer should scan the string content.

You are right in that there is no reason to have Elixir's uppercase/lowercase convention for escape code and interpolation scanning in Erlang, other than for having similar look and feel between the languages, when possible.  

Another approach could be to interpret the prefix as characters and have the lexer recognize some character for escape codes, interpolation, and whatnot...  But if we bundle escape codes + interpolation in the same flag, then the uppercase/lowercase convention works.

eeps/eep-0065.md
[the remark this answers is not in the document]
RaimoNiskanen: > After all, the `r` sigil uses different escape sequences than s.

Oh, does it?  Interesting! What are the differences / where is that documented?

And, yes, we need to decide on the matter of extensible sigils, and then how the lexer and the parser should handle them...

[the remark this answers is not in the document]
josevalim: Here is an example: `\s` in a string is a whitespace. `\s` in a regex means any whitespace character (\n, \t, \s). Not only that, when you see `\[` in a string, you silently discard the `\`. However, in a regex, you want to keep the escape character around and send it all the way down to the regex engine.

Here is the unescape map for regexes from Elixir:

https://github.com/elixir-lang/elixir/blob/bc3c915563a59c40c476fd2d512d43ccdac793e6/lib/elixir/lib/regex.ex#L857

Here is the default unescape map for strings:

https://github.com/elixir-lang/elixir/blob/bc3c915563a59c40c476fd2d512d43ccdac793e6/lib/elixir/src/elixir_interpolation.erl#L251-L265

Returning false means "leave the escape character in, do not remove it".

asabil:
On the topic of extensibility, would it make sense to consider introducing `string_transform` similar to `parse_transform`?

RaimoNiskanen:
@asabil: String transformation can sure be done in a parse transform.  Can you sketch how you think a `string_transform` should work?

RaimoNiskanen:
I rewrote quite a bit, new proposal:
* Sigil type is a name
* String delimiters define verbatim
* Increase the set of delimiters
* Talk down general expression interpolation in strings

eeps/eep-0066.md
```diff
@@ -0,0 +1,343 @@
...
+
+Between the start and end [String Delimiters][], all characters
+are string content.
+
+The delimiters «`'`» and «`"`» work as they do for quoted atoms
+and regular strings which allow escape sequences in the string content.
+
+All other delimiters do not allow escape sequences; all characters
+are verbatim up to the end delimiter, and there is no way
+to escape the end delimiter.
```
josevalim, line 202: How would this work for regular expressions? The issue with regular expressions is that there are two escapes you can perform: at the tokenizer level and at the sigil level.

You may say that `~r/f\no/` won't escape a newline on the tokenizer but you still want `~r/\[\d\]$/` to match on a string such as "foo[0]" and, according to my interpretation of escape sequences, `\[, `\d`, and `\]` are all escape sequences (or at least the first and last one) and all of them are handled by the regex compiler. This means the sentence above in the EEP is not actually true.

If you intend to say it disables the tokenizer escaping but keeps the sigil one, then there is the issue that `~r"\s"` will behave slightly differently from `~r/\s/`, because one escape `\s` in the tokenizer (to literally mean whitespace) and the other inside the regex means any whitespace, and this will likely be a big pitfall.

You could also say that it will disable all escape characters, but then `~r/anything/` is not useful if you can't use any of regex escapes.

I may be missing other options, but none of the above sound optimal to me.

---

I do realize this is pretty much an issue in Elixir too! After all, `~r/\[\d\]$/` and `~R/\[\d\]$/` behave the same in Elixir, so we are actually lying if we say that uppercase disables escaping (with ~R being the counter-example).

However, Elixir has a way out, we can just deprecate `~R` altogether. This EEP, as written, does not have this option, because the delimiter is the one that controls escaping. And, if one of the main points for introducing `/` in Erlang is to work with regexes, then it must at least allow the regex escapes to be useful.

In my humble opinion, keeping the control of escaping as part of the name may be more flexible in the long term (either as casing or as a convention, such as `us` for unescaped strings).

RaimoNiskanen: I do not quite understand what you mean by sigil level excapes...

My intention was to say that `~r/f\no\[\d\]$/` will pass all characters verbatim i.e  `f\no\[\d\]$` to the regexp compiler.  But `~r"f\n\o\[\d\]$"` will pass `[102,10,111,91,127,93,36]` to the regexp compiler since it follows normal Erlang string escape rules.
To get a newline (10) between the `f` and the `o` you can use `~r"f\no\\[\\d\\]$"`, but I suppose regular expressions handle `\n` as a newline so the first one, `~r/f\no\[\d\]$/` is probably the right one.

This suggestion explores a different kind of consistency, that the `"` (the `"""`) and the `'` delimiters behave consistently (as before) in the language, and the new delimiters are around verbatim sigil strings.

So the issue that `~r"\s"` would behave differently from `~r/\s/` is intentional.

The question if if having the delimiters determine tokenizer escaping is a good or a bad idea...

You said:
> because the first escapes `\s` in the tokenizer (to literally mean whitespace) and the second inside the regex means any whitespace, and this will likely be a big source of confusion.

_("literally mean whitespace should be "literally mean the space character")_
Do you think this would be a big source of confusion when being used to the Elixir convention, or that having this difference between the language would continue to be confusing.

I am trying to gather arguments for and against delimiter based tokenizer escaping.  I also have a vague feeling like you that
> more flexible in the long term (either as casing or as a convention, such as us for unescaped strings), rather than as part of the delimiters.

But I cannot really put my finger on it...

Right now, if not looking to hard at Elixir, I think that having the `"`, `"""` and `'` delimiters behave as before and having new delimiters for verbatim strings could be kind of simple and consistent.

josevalim: > I do not quite understand what you mean by sigil level excapes...

My issue is that this sentence: "All other delimiters do not allow escape sequences" is incorrect for regular expressions. You could say "All other delimiters do not allow escape sequence except by the escapes used by the sigil itself" or that "we pass options verbatim to the regex compiler" but that's an implementation detail. I'd say the user should not know or care where the escaping is happening.

So, from the user perspective, if `~r/\-/` matches "foo-bar", then `\` is being used as escape character, and therefore the sentence is false. If I were to take the sentence literally, I would expect `~r/\-/` to match only strings that have both `\` and `-` in sequence. But then the `~r` sigil with `/` as delimiters is hardly useful.

>  Do you think this would be a big source of confusion when being used to the Elixir convention, or that having this difference between the language would continue to be confusing.

I would say both. But, as you said, it is a gut feeling. If I try to quantify it a bit, here are some discussion points:

  * which one the notations below better indicates to the user they are different:
    * Uppercase style: `~r/foo/` vs `~R/foo/`
    * Delimiter style: `~r"foo"` vs `~r/foo/`

  * Given the `\s` pitfall, my concern is that, by using delimiters to denote escaping, you will most likely end-up with community rules such as "do not use quotes as delimiters for regular expressions". Then we should wonder if it is not best to remove the possibility altogether. For example, you could raise if someone is using quotes with `~r` at the tokenizer/parser level (meaning different sigils will have different delimiters).

As I mentioned before, the `~R` sigil in Elixir was also inconsistent here, so I was wondering why this never came up before. It turns out that, after grepping my disk across 87 Elixir projects plus all of their dependencies, only two dependencies used `~R`, one of them being to intentionally escape an interpolation, and the other could be replaced by `~r` as is.

---

One possible suggestion here is to reduce the initial scope of the EEP. What we know is that you want to introduce `~c`, `~s`, and `~r` and all of them use escape characters. So you could introduce the three sigils, using only quotes as delimiters, giving you the flexibility to either add multiple delimiters or add lowercase/uppercase conventions in the future. Even without the ability to turn off escaping, I would say the EEP would bring drastic improvements:

  * `~s"José"` beats `<<"José"/utf8>>` (IMO)
  * `~r"[\w\-]\[\d+\]"` beats `re:compile("[\\w\\-]\\[\\d+\\]")` (this one should not be debatable :P)

After all, adding language features is much easier than removing. And, if you were to add new features in the future, you would probably be able to do so based on usage and feedback within the Erlang community?

essen: > One possible suggestion here is to reduce the initial scope of the EEP. What we know is that you want to introduce `~c`, `~s`, and `~r` and all of them use escape characters. So you could introduce the three sigils, using only quotes as delimiters, giving you the flexibility to either add multiple delimiters or add lowercase/uppercase conventions in the future. Even without the ability to turn off escaping, I would say the EEP would bring drastic improvements:
> 
> * `~s"José"` beats `<<"José"/utf8>>` (IMO)
> * `~r"[\w\-]\[\d+\]"` beats `re:compile("[\\w\\-]\\[\\d+\\]")` (this one should not be debatable :P)
> 
> After all, adding language features is much easier than removing. And, if you were to add new features in the future, you would probably be able to do so based on usage and feedback within the Erlang community?

Strong +1 on wanting to limit the scope, introduce 2 or 3 sigils, keep them verbatim only, wait and see what the response is. I'm not even sure ~c should be there but I guess people still use them strings.

RaimoNiskanen: @josevalim wrote:
>  or that "we pass characters verbatim to the regex compiler" but that's an implementation detail. I'd say the user should not know or care where the escaping is happening.

Ah!  I did not think of that perspective and have mentally placed the regexp compiler as obviously outside this context.  Important point!

But I think the user must be aware of which escape rules that are used for which sigils.

> One possible suggestion here is to reduce the initial scope of the EEP. What we know is that you want to introduce `~c`, `~s`, and `~r` and all of them use escape characters. So you could introduce the three sigils, using only quotes as delimiters ...

From the sigil's point of view, the `~c` and `~s` sigils should handle escape sequences in the tokenizer, but the `~r` sigil needs the tokenizer to be as verbatim as possible, passing the implementation of escape sequences to the regexp engine that we have chosen.

So this EEP needs to decide on how the tokenizer should know whether to do Erlang's escape sequence expansion or not, and it needs to decide on exactly how verbatim the tokenizer should be for regular expressions.

Other than that it sure sounds like a good idea to start minimal and expand later.  But the start point mustn't paint us into a bad corner...

RaimoNiskanen:
I stepped back to Sigil Type based verbatim:ness and reduced the scope of Sigil Types and String Delimiters

eeps/eep-0066.md
```diff
@@ -0,0 +1,340 @@
...
+  on the [String Content][].  The [String Delimiters][]
+  and escape characters work as for regular strings,
+  triple-quoted strings, or quoted atoms in Erlang.
+
+  So «`~"abc\d"`» is equivalent to «`<<"abc\d"/utf8>>`», and
+  «`~'abc"d'`» is equivalent to «`<<"abc\"d"/utf8>>`».
+
+* «`r`»: regular expression.
+
+  Creates a term `{RE::unicode:charlist(),Flags::[unicode:latin1_char()]}`
```
josevalim, line 181: Perhaps the topic of another conversation, but Elixir compiles the regular expression at compilation time. We have to store the endianess and re version, and recompile on the fly if those change, but Erlang could consider doing the same.

eeps/eep-0066.md
```diff
@@ -0,0 +1,340 @@
...
+
+Immediately following the [Sigil Type][] is the string start delimiter.
+A specific start delimiter character has a corresponding
+end delimiter character.
+
+These are the start delimiter characters and they have themselves
+as end delimiter: single quote «`'`», and double quote «`"`».
+
+Triple-quote delimiters are also allowed, that is; a sequence of
+3 or more double quote «`"`» characters as described in [EEP 64][].
```
josevalim, line 208: I have re-read EEP 64 and it may be worth considering the meaning of verbatim strings with sigils. If `"""` are intended to be verbatim, then should `~"""` also be verbatim? If so, that would be the default mechanism of disabling escaping. If not, it is probably worth adding a line to this paragraph clarifying the relationship? :)

josevalim:
I like the reduced scope of the proposal a lot. I have also added a comment about verbatim strings and sigils. Btw, feel free to mark my previous comments as resolved, unfortunately I cannot. :)

eeps/eep-0066.md
RaimoNiskanen: I will add a chapter about the reasoning behind this sigil output type.  When to compile an RE is part of that.

So far, to my knowledge, we don't have anything cached for a specific platform in the .beam files.  We have talked about compiling an RE at load time.  That is also when the Just In Time compiler runs.

eeps/eep-0066.md
RaimoNiskanen: My intention is, but since I reduced the scope of the EEP it got removed, that `~s"` should do escaping, and therefore so should `~s"""` since the `s` is explicit.  It would be confusing if `~s` behaved differently for `"` and for `"""`.

Then, also, `~S` should be verbatim, both for `"` and for `"""`.

I think an important use case would be making a UTF-8 binary out of a `"` or a `"""` string therefore it should be simple, as in just adding a `~` sigil to it.  So the proposal is that `~"` means `~s"` and `~"""` means `~S"""`.

The other possibility would be that `~` is a shorthand for `~s`, which would be implicit and contradict the other visible cue: `"` or `"""`, so it is hopefully intuitive that the quote decides.

josevalim: > I think an important use case would be making a UTF-8 binary out of a " or a """ string therefore it should be simple, as in just adding a ~ sigil to it. So the proposal is that ~" means ~s" and ~""" means ~S""".

Thanks. I think adding something along those lines would perfectly clarify it.

RaimoNiskanen:
I have compared the vanilla sigil `~` with future `~s` and `~S`, and explained why I think the `~r` sigil should be an uncompiled regular expression.

asabil:
> @asabil: String transformation can sure be done in a parse transform. Can you sketch how you think a `string_transform` should work?

I haven't really given it much thought, but the idea would be to allow something like the code below.

```erlang
-module(st_test).
-compile({string_transform, sql, sql_transform, []}).

-export([
    test/1
]).


test(Table, ID) ->
    ~sql~"SELECT FROM $[Table] WHERE id = ${ID}".
```

```erlang
-module(sql_transform).
-export([
    string_transform/2
]).

-include_lib("syntax_tools/include/merl.hrl").

string_transform(StringNode, _Options) ->
    parse(erl_syntax:string_value(StringNode)),
    {Query, Params} = lists:foldl(fun
        ({param, #{name := ParamName}}, {QueryAcc, ParamAcc}) ->
            ParamValue = merl:var(binary_to_atom(ParamName)),
            {[?Q("<<\"?\">>") | QueryAcc], [?Q("_@ParamValue") | ParamAcc]};
        ({ident, #{name := IdentName}}, {QueryAcc, ParamAcc}) ->
            IdentValue = merl:var(binary_to_atom(ParamName)),
            {[?Q("sql:quote(_@IdentValue)") | QueryAcc], ParamAcc};
        (Node, {QueryAcc, ParamAcc}) when is_binary(Node) ->
            {[?Q("_@Node@") | QueryAcc], ParamAcc}
    end, Template),
    erl_syntax:tuple([
        erl_syntax:list(lists:reverse(Query)),
        erl_syntax:list(lists:reverse(Params))
    ]).


-spec parse(string()) -> [Token] where
     Token :: binary() | {param, #{name := binary()}} | {ident, #{name := binary()}}].
parse(String) ->
   []. %% TODO: implement parsing

```

Resulting in

```erlang
1> st_test:test(user, 1).
{[<<"SELECT FROM ">>, <<"\"user\"">>, <<" WHERE id = ?">>], [1]}.
```

It's maybe nothing more than a simplified way of implementing a parse transform. Not sure how useful that would be?

eeps/eep-0066.md
```diff
@@ -0,0 +1,432 @@
...
+  Between triple-quote delimiters according to [EEP 64][]
+  there is no end delimiter character to escape.
+
+### String Delimiters
+
+Immediately following the [Sigil Type][] is the string start delimiter.
+A specific start delimiter character has a corresponding
+end delimiter character.
+
+These are the start delimiter characters and they have themselves
```
kikofernandez, line 229: I understand correctly that we do not have plans for supporting delimiters like Elixir, which could be:
```
~r/hello/
~r|hello|
~r"hello"
~r'hello'
~r(hello)
~r[hello]
~r{hello}
~r<hello>
```
https://elixir-lang.org/getting-started/sigils.html

If so, I agree that it is better to start small and do incremental changes / additions if needed in the future.

eeps/eep-0066.md
```diff
@@ -0,0 +1,432 @@
...
+### String Content
+
+Between the start and end [String Delimiters][], all characters
+are string content.
+
+In a triple-quoted string all characters are verbatim, but stripping
+of indentation and leading and trailing newline is done as usual
+as described in [EEP 64][].
+
+In a string with single character [String Delimiters][],
```
kikofernandez, line 244: This is consistent with triple-quoted code and single-quoted code
To me, it is intuitive

eeps/eep-0066.md
```diff
@@ -0,0 +1,432 @@
...
+normal Erlang escape sequences prefixed with «`\`» are honoured,
+as usual for regular Erlang strings and quoted atoms
+
+A specific [Sigil Type][] can have it's own character escaping rules,
```
kikofernandez, line 248: it's -> its

eeps/eep-0066.md
```diff
@@ -0,0 +1,432 @@
...
+The Sigil Suffix consists, as the [Sigil Type][] in the [Sigil][] prefix,
+of name characters.
+
+The Sigil Suffix may indicate how to interpret the String Content,
+for a specific [Sigil Type][].  For example; for the «`~r`» [Sigil][]
+(regular expression), the Sigil Suffix is interpreted as short form
+compile options such as «`i`» that makes the regular expression
+character case insensitive.
+
+Things that may have to be performed by the tokenizer, such as
```
kikofernandez, line 265: Yes, the tokenizer should not be influenced by how one appends a suffix.

eeps/eep-0066.md
```diff
@@ -0,0 +1,432 @@
...
+
+With this representation, the `re` module can be augmented
+with functions that accept this tuple format.  These functions
+are `re:compile/1,2`, `re:replace/3,4` `re:run/2,3`,
+and `re:split/2,3`  Translation of the compile flag characters
+into `re:compile_option()`s should be done by these functions.
+
+Example:
+
+    1> re:run("ABC123", ~r"abc\d+"i, [{capture,first,list}]).
```
kikofernandez, line 286: This example considers that the `re:run` function has been augmented.
For the purpose of clarity, could we add what the sigil desugars into and mention specifically the change needed in `re:run`? We can end by saying that those changes apply to `re:compile` et al.

eeps/eep-0066.md
```diff
@@ -0,0 +1,432 @@
...
+    {match,["ABC123"]}
+
+Since the [Sigil][] value represents an uncompiled regular expression,
+the user can choose when to compile it with `re:compile/1,2`,
+or use it directly in e.g `re:run/2,3`.
+
+It is possible to implement an optimization to make the compiler
```
kikofernandez, line 293: This is a nice addition but not a must for the EEP.
But good that it is documented.

eeps/eep-0066.md
```diff
@@ -0,0 +1,432 @@
...
+pre-compilation, then such an optimization is safe.
+
+#### Compiled Regular Expression
+
+Another possibility would be that the value of a regular expression
+[Sigil][] is a compiled regular expression; the `re:mp()` type.
+
+Then it can be used as above, *except* as an argument to
+`re:compile/1,2`.  Pre-compilation would be a hard requirement
+since the running Erlang code *must* see a compiled regulare expression.
```
kikofernandez, line 314: regulare -> regular

eeps/eep-0066.md
```diff
@@ -0,0 +1,432 @@
...
+And we would still have to decide on another sigil type to be used
+in `re:compile/1,2` that is syntactic sugar for an *uncompiled*
+regular expression.  Without that a `~S` sigil could be used
+but that won't have the compile flags as suffix so those flags
+cannot be given in the same way for compiled vs. uncompiled
+regular expressions.
+
+#### Therefore Uncompiled
+
+Since we in any case need a [Sigil][] that is syntactic sugar for
```
kikofernandez, line 325: Yes, start small and increment

kikofernandez (approved):
I think the EEP is quite complete and thorough.
I approve pending fixes in misspellings.

Good work @RaimoNiskanen

eeps/eep-0066.md
jhogberg: I don't think compiling REs at load-time is a good idea: if we abandon PCRE we'd start getting mysterious failures at load-time that may be difficult for the user to pinpoint. `on_load` also has nasty implications for code loading which, among other things, prevent loading modules in parallel.

I think it's better to leave it to the user, there's little stopping them from compiling the REs in an initialization phase and putting them in `persistent_term`.

bjorng (approved):
Sounds reasonable to me. I like the use a unique prefix character. That makes it easier to see where the sigil begins, and should also leave us with flexibility for extension in the future.

eeps/eep-0066.md
```diff
@@ -0,0 +1,432 @@
...
+  there is no end delimiter character to escape.
+
+### String Delimiters
+
+Immediately following the [Sigil Type][] is the string start delimiter.
+A specific start delimiter character has a corresponding
+end delimiter character.
+
+These are the start delimiter characters and they have themselves
+as end delimiter: single quote «`'`», and double quote «`"`».
```
bjorng, line 230:
```suggestion
as end delimiter: single quote «`'`» and double quote «`"`».
```

eeps/eep-0066.md
RaimoNiskanen: It was in the EEP up to just recently.  It is a possible future augmentation.  Therefore the EEP still talks about start and end delimiter characters instead of just delimiter character.

By reducing to just two delimiters I also made the choice that the sigil name sets escape rules, not the chosen delimiters.  Like in Elixir.

RaimoNiskanen: @josevalim said that he hasn't seen much use for other delimiters than `"`.  We have `'` for quoted atoms, so I kept them too.

eeps/eep-0066.md
RaimoNiskanen: Isn't that explained in the paragraph just above this example?

eeps/eep-0066.md
bjorng: > @josevalim said that he hasn't seen much use for other delimiters than `"`. We have `'` for quoted atoms, so I kept them too.

`\` does seem to be used for regular expressions. But strangely enough for all. Grepping in the `elixir` repository:

```
% git grep "~r/" | wc
     414    3241   44627
% git grep "~r""" | wc 
    1028    9004  117317
```

(I am surprised that `~r/` is not used for the majority of regexes.)

eeps/eep-0066.md
```diff
@@ -0,0 +1,438 @@
...
+  the [end delimiter][] character can be escaped with a «`\`» character.
+  Between triple-quote delimiters according to [EEP 64][]
+  there is no end delimiter character to escape.
+
+  The main advantage of a regular expression [Sigil][] is to avoid
+  the additional escaping of «'\'» that regular erlang strings add.
+
+  Today: `re:run(Subject, "^[ \\t]*\\[a-z]*\\\\s+", [caseless,unicode])`
+
+  Sigil: `re:run(Subject, ~r"^[ \t]*\[a-z]*\\s+"iu)`
```
jhogberg, line 228: The more I think about it the less I like the idea of having a specific sigil for regular expressions, what is the main benefit over `re:run(Subject, ~S"^[ \t]*\[a-z]*\\s+", [i, u])`?

michalmuskala: Yes, I think it would be valuable if we could compile them at load-time. At time point the value is relatively limited.
Unless the representation of `~r` is left unspecified, such that it would be backwards-compatible to later compile them at load-time in a subsequent release.

RaimoNiskanen: I'd say possible `re` integration, future pre-compilation, and the possibility to handle the regular expression and its compilation flags as one unit (+ less punctuation characters for the compile flags).

RaimoNiskanen: If we would take an intermediate step to not add `~r` now, then people will start using `~S` for regular expressions, and then want to switch to using `~r` when that is introduced.

I don't know if it will be annoying to have stepped intermediate step...

essen: I think `~r` would make more sense if it was tied to an operator like the Perl match operator. Maybe use multiple sigils, like `Subject <> ~m"..."` to match (or `~m/.../`), `Subject <> ~s".../..."` to replace. The pre compilation alone that `~r` could provide is not super interesting.

RaimoNiskanen: I do not think we want to tie regular expressions into the base language that hard, as an operator would do. Erlang should not try to be Perl.

If we would have a match operator, then there would be no possibility to e.g strip the `re` module from Stdlib.  It would irreversibly be a part of the Erlang language. I think it should be a library.  But some language support like this would be more OK.

essen: The pre compilation discussed here would also tie the Erlang language to a regular expression implementation though. I do not think an operator is too different. Just have to devise a way to make the implementation pluggable instead of relying directly on `re`.

The operator doesn't have to specify the way the regular expression implementation behaves, just what happens when there is/isn't a match. Then users can refer to the implementation they use for the details of the syntax and behavior.

I'm not saying this should be just like Perl, as Perl goes far beyond this and not in a good way IMO. But I don't see much point in having `~r` if it's going to either be like `~S` or just provides some pre compilation magic on top.

josevalim: > Maybe use multiple sigils, like Subject <> ~m"..." to match (or ~m/.../), Subject <> ~s".../..." to replace. 

I am not the biggest fan of making the result of the operation based on the sigil. Maybe `Subject <> ~m"foo"` means match, maybe `Subject <> ~s"foo/bar"` means replace, but you can also write `Subject <> AnotherVar`, and now I have no idea what it is going to return without knowing the exact type of `AnotherVar`.

I would also say that, from experience, match itself is not terribly useful. In Perl, `=~` sets magic variables with the regex captures, so it works both as `re:run/2` or also a quick check to see if something matched or not.

I believe no one is advocating for magic variables in Erlang, so the only option is to use `=~` (or similar) to check if something matched or not. We do have this operator in Elixir and, to be honest, the operator is almost exclusively used in tests, and that's it. Because in actual code, you most often don't want to just check if a regex matches, but also extract something from it. If that's the primary benefit, does it justify the changes to the language?

essen: In actual code I almost never see or use regular expressions, so I think the question applies to any language feature related to them. Perhaps the benefits of `~r` or equivalent are just too small to matter.

About the operator obviously that wasn't a fully thought out idea, and sigils might not be the right way to indicate match/replace, definitely. But I didn't think of `Subject <> ~m"foo"` as match-only but as something that would return the captures if any were defined.

Anyway details aside I think that my point that `~r` on its own not being super useful still stands. If the language allows defining regular expressions it should also allow using them directly otherwise they're no better than regular strings.

josevalim: That's a chicken-egg issue though. Regexes are very awkward to write today, so it is understandable there are rarely used. I would say we should at least make them easier to write.

essen: I'm not disagreeing with `~S` which would make them easy to write already. The verbatim stuff is great. I'm just saying if you introduce something specific to regular expressions it has to either come with the relevant features or that those features are intended to be introduced at a later time. Stopping at `~S` + maybe a pre compilation later is not interesting IMO.

essen:
How do sigils play with concatenation? Can you do `~r"\W*((?i)" ++ escape(Word) ++ ~r"(?-i))\W*"`? How does that play with potential features like pre compiling the regexps?

eeps/eep-0066.md
RaimoNiskanen: @essen wrote:
> The pre compilation discussed here would also tie the Erlang language to a regular expression implementation though. I do not think an operator is too different. Just have to devise a way to make the implementation pluggable instead of relying directly on re.

An operator would require that there *is* an regular expression engine in the system or else the syntax does not make sense.

The current suggestion is that `~r"abc\d"ef` translates to `{re,"abc\\d","ef"}`, which is a pluggable interface, and fairly general.  In most other interfaces I know of there is the expression and the compile flags, both as strings.  Then a regular expression module should accept such a term for the combined regular expression and its compile flags.

If there is such a module, the code loader can be made to pre-compile regular expression literals used to call functions in the regular expression module.

> But I don't see much point in having ~r if it's going to either be like ~S or just provides some pre compilation magic on top.

The other benefit is to be able to combine the regular expression and its compile flags into one unit, in a way familiar to other regular expression interfaces such as sed, awk, perl, ...

RaimoNiskanen:
> How do sigils play with concatenation? Can you do `~r"\W*((?i)" ++ escape(Word) ++ ~r"(?-i))\W*"`? 

If `~r` produces a tuple, then that would not work.  But for sigils that produce a string it would.  Not for `~S`, though since it produces a binary.  But `[~S"\W*((?i)", escape(Word), ~S"(?-i))\W*"]` would be an `unicode:charlist()` with the right content.

Earlier, when i thought about concatenating sigils it seemed to me that `~x"abc"y ~x"def"y` might be possible to combine into `~x"abcdef"y`, just like strings are concatenated in the parser, but I guess that for some other future sigil even that is not safe.  And with expressions in between the parser cannot do anything.

A regular expression sigil with variable interpolation could be useful in this case, something like: `~rv"\W*((?i)~{Word/escape}(?-i))\W*"`.
That could not be pre-compiled at load time since it cannot be a literal.  I guess it could expand to `{re,"\\W*((?i)"++escape(Word)++"(?-i))\\W*",""}` or `{re,[<<"\\W*((?i)"/utf8>>, escape(Word), <<"(?-i))\\W"/utf8>>], ""}`

essen:
> > How do sigils play with concatenation? Can you do `~r"\W*((?i)" ++ escape(Word) ++ ~r"(?-i))\W*"`?
> 
> If `~r` produces a tuple, then that would not work. But for sigils that produce a string it would. Not for `~S`, though since it produces a binary. But `[~S"\W*((?i)", escape(Word), ~S"(?-i))\W*"]` would be an `unicode:charlist()` with the right content.

Sounds like sigils are all inconsistent with each other and provide us with many pitfalls.

> Earlier, when i thought about concatenating sigils it seemed to me that `~x"abc"y ~x"def"y` might be possible to combine into `~x"abcdef"y`, just like strings are concatenated in the parser, but I guess that for some other future sigil even that is not safe. And with expressions in between the parser cannot do anything.

It's fine if it's not the parser doing the work, it can be at runtime. But not being able to concatenate `~r` sigils is not intuitive since what we see is a string (in the `string` module sense, so `string:concat` should work), while under the hood it's a tuple. I expect a few beginner errors.

> A regular expression sigil with variable interpolation could be useful in this case, something like: `~rv"\W*((?i)~{Word/escape}(?-i))\W*"`. That could not be pre-compiled at load time since it cannot be a literal. I guess it could expand to `{re,"\\W*((?i)"++escape(Word)++"(?-i))\\W*",""}` or `{re,[<<"\\W*((?i)"/utf8>>, escape(Word), <<"(?-i))\\W"/utf8>>], ""}`

Not a fan of interpolation at all, coming from PHP long ago. Even more so in Erlang with all the `Var1`, `Var2`, `Var3` shenanigans. That aside, variable interpolation will inevitably lead to some expressions needing additional escaping. But at the same time, if some expressions are written using `~r` and others using `~S` it's also a problem: now you'll have half the expressions using `~r` and half using `~S`. But AFAICT that's what it's going to be initially since there won't be interpolation. Not to mention the code debt.

Instead of straight up interpolation I would much prefer that you define the regular expression as `~rv"\W*((?i)~{word/escape}(?-i))\W*"` but then have to provide `word` as argument to the function from `re` (or other module) as a map `#{word => "foo"}`. A bit like SQL parameterized queries, with all the pro and cons. This type of input also allows engines to potentially pre compile even if they don't have the variable yet. But this does mean some expressions will need additional escaping.

josevalim:
I don't think it is adequate to say `~r"foo"` is a string and it should behave as a string. In the same way that, `'foo'` is not a string and `'foo' ++ 'bar'` is not a valid operation. There may be some initial confusion but those should be quickly addressed by stating that different sigils will yield different data types with different operations for them.

> Instead of straight up interpolation I would much prefer that you define the regular expression as ~rv"\W*((?i)~{word/escape}(?-i))\W*" but then have to provide word as argument to the function from re (or other module) as a map #{word => "foo"}. 

That's how Python and (and somewhat Java) has historically done this but both are migrating to interpolation based syntax:

 * https://openjdk.org/jeps/430
 * https://peps.python.org/pep-0502/

I am not advocating in favor of one or the other, but one could look at the on-going discussions for rationale, motivations, etc. Generally speaking there seems to be a push/preference towards interpolation.

essen:
> I don't think it is adequate to say `~r"foo"` is a string and it should behave as a string. In the same way that, `'foo'` is not a string and `'foo' ++ 'bar'` is not a valid operation. There may be some initial confusion but those should be quickly addressed by stating that different sigils will yield different data types with different operations for them.

It's visually a string with `~r` in front of it. A single character determines whether you can concatenate or not. And initially it would be the only sigil that you can't handle as a string. It would be more obvious if it was `~r/.../` for example. I don't expect a lot of confusion because regular expressions are rarely used, but this will crash a few programs.

> I am not advocating in favor of one or the other, but one could look at the on-going discussions for rationale, motivations, etc. Generally speaking there seems to be a push/preference towards interpolation.

Yes they're generally convenient, especially when it comes to Web development. I suppose I would be fine with them if there was a switch to keep them disabled in 95% of modules. Then in the remaining 5% I can be extra careful about the environment of the functions using interpolation. Otherwise it's just too risky compared to a string templating approach. We could have both too, of course.

RaimoNiskanen:
> Sounds like sigils are all inconsistent with each other and provide us with many pitfalls.

The idea with sigils, as proposed, I'd say, is to transform a string into some term according to some rules.  The sigil type defines the created type.

`~S` creates a binary so it cannot be treated as a string, e.g. `++` does not work.
`~s` -"-
`~C` explicitly creates a character list so *that* is a string, but that is an exception.
`~c` -"-
`~r` is suggested to create a 3-tuple.

And `~S"abc"` might look like a string with an "operator" `~S` prepended, such as `not"abc"` which is valid syntax and could be a boolean (but it badargs at runtime).  My point is that it should not be surprising that an operator may produce a new type.

RaimoNiskanen:
I changed to propose `~b` and `~B` to create binaries, and `~s` and `~S` to produce strings.

eeps/eep-0066.md
```diff
@@ -0,0 +1,496 @@
...
+
+  In Elixir this corresponds to the «`~c`» sigil, a [charlist][5].
+
+* «`S`»: `string()`, verbatim.
+
+  Creates a Unicode codepoint list, with verbatim string content
+  in that only the [end delimiter][] character can be escaped
+  with a «`\`» character.
+
+  In Elixir this corresponds to the «`~C`» sigil, a [charlist][5].
```
josevalim, line 217: A string in Erlang, according to the `string()` module, is any `unicode:chardata()`. According to types, `string()` is `[char()]`. For this reason, I would go with `~c` to avoid any source of confusion, but it is not a strong opinion. :)

RaimoNiskanen: I was thinking about the built in type `string()`, but OK, there is a possible source of confusion here...

RaimoNiskanen:
I went back to the original set of delimiters, after deciding that it is the sigil type only that affects the tokenizer.  For regular expressions it sometimes looks good with `/` or `|`, for example.

RaimoNiskanen:
An internal meeting disagreed with me, so for now we only keep the `"` string delimiters (and `"""`), and also postpone the `~r` sigil.

PR #54, opened 2023-09-28
README: Fix links to EEP 0

juhlig:


RaimoNiskanen (commented):
The old link points to our download server, on which we apparently stopped serving this directory.
So this link redirects to the web server at www.erlang.org that for the /eeps/ directory auto appends .html.  I think we should not rely on that behaviour so the new link should be to `http://erlang.org/eeps/eep-0000.html`.
Then the web server should stop removing .html from the links it translates within EEP 0.
I think it should not be hidden that there is the source format .md that is translated to the view format .html, and that the web server should serve both, explicitly.

PR #58, opened 2024-01-16
EEP: internal export

MarkoMin:
EEP for introducing the concept of *internal exports*. For details, look in the EEP itself and in the PR of reference implementation.

P.S. Do I need to send it on the mailing list?

### DIFF
--- a/eeps/eep-0067.md
+++ b/eeps/eep-0067.md
@@ -1,0 +1,237 @@
+    Author: Marko Minđek <marko.mindek(at)invariant(dot)hr>,
+            Karlo Nikšić <kuna.prime(at)invariant(dot)hr>
+    Status: Draft
+    Type: Standards Track
+    Created: 02-Jan-2024
+    Erlang-Version: OTP-27
+    Post-History:
+****
+EEP XXX: Internal export
+----
+
+Abstract
+========
+
+This EEP introduces a new directive called `internal_export` which
+enables semantic separation of function exports.  This EEP is mostly
+inspired by [EEP 5][].
+
+Rationale
+=========
+
+Erlang application API is an ambigous term.  In theory, it is a set
+of module APIs (exported functions).  In practice, it is a set of
+modules' APIs but without modules considered internal, undocumented
+exports, callback implementation, etc.  While reading the docs,
+you can conclude what is application API, but while reading the
+source, it's not so straightforward.
+
+Not all exports are semantically the same.  A function can be exported
+to be part of the application API, to implement a callback, to test
+some code, or to allow modules within the application to use it.
+Currently, from the user perspective, they are all the same.  There
+is a convention not to use undocumented exports, but there is no way
+to enforce that convention in a codebase.
+
+The goal of this EEP is to give a mechanism to separate application API
+exports from other exports.
+
+A few terms are introduced for convenience:
+
+- internally exported function - a function marked with
+`internal_export` attribute
+- internal module - module containing only internally exported functions
+- application dependencies - a set of applications found in `applications`,
+`include_applications` and `optional_applications` properties in
+`.app` file (and their dependencies, recursively)
+- export scope - a set of modules that can *legally* access the
+exported function
+
+Internally exported functions are still globally exported, but
+application users are discouraged from using them.  The mechanism is similar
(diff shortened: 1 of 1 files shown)


josevalim:
Given this EEP proposes to add a new module attribute that is considered by xref, it looks quite straightforward to me compared to the previous EEP, which wanted to bring this knowledge to the compiler and the loader, and therefore had wider ranging implications. With this in mind, I wonder if internal_export is the best name or if the attribute should have xref in its name. The downside is that internal exports would have to be listed twice, one as a regular export, and another, with its visibility rule.

Another question brought up by the EEP is if exporting to applications is indeed the best practice. For those working on large applications (a.k.a. monoliths), You may still want to introduce some sort of boundary between different parts of our system without breaking into smaller applications. A potential alternative here is to allow developers to define the visibility using regular expressions, prefixes/suffixes, or combination of rules, such as:

  -xref_app_export {some_app, […]}.
  -xref_mod_export {“my_app_*”, […]}.
  
I am not proposing to add all those possibilities upfront, but consider them in case more flexibility is required in the future.

MarkoMin:
> Given this EEP proposes to add a new module attribute that is considered by xref, it looks quite straightforward to me compared to the previous EEP, which wanted to bring this knowledge to the compiler and the loader, and therefore had wider ranging implications. 

It's currently considered only by `xref`, but it's not limited to. Documentation tools could use it, shell could provide better completion based on it, language servers, . I'm not sure that `internal_export` is the best name, but I'm sure I don't want to bind it directly to `xref`.
Furthermore, the plan for the future is to have `internal_export_type`, because a lot of types are used only from within the app itself - but those are checked by dialyzer (or any other tool). Notice also that nothing restricts us from doing the check in the loader/runtime in the far future (not saying that we should add it, but if we do, it could be on top of the current implementation)



> With this in mind, I wonder if internal_export is the best name or if the attribute should have xref in its name. The downside is that internal exports would have to be listed twice, one as a regular export, and another, with its visibility rule.

Yes, but I solved this with `internal_export` feature with expands `-internal_export([f/a, ...]).` to both internal export and export. There are probably other ways too.


> Another question brought up by the EEP is if exporting to applications is indeed the best practice. For those working on large applications (a.k.a. monoliths), You may still want to introduce some sort of boundary between different parts of our system without breaking into smaller applications. A potential alternative here is to allow developers to define the visibility using regular expressions, prefixes/suffixes, or combination of rules, such as:

In the EEP I assumed that *misusage*  comes from the app's users, not from within the app itself. We could possibly extend the attribute to take 2 arguments, where second one is the list of modules, that would enable more intra-app control. In which use case you want to export something to another app (except for callbacks, which are not statically called anyways)?

kikofernandez:
@MarkoMin Thanks for your contribution! We will discuss this EEP at the end of week 5, so you should hear from us at the end of week 5 or on week 6.

In the meantime, would it be possible for you to fix the markdown syntax as per the EEP guidelines [[1](https://www.erlang.org/eeps/eep-0001), [2](https://www.erlang.org/eeps/eep-0033)] so that we can add this EEP to the list of proposed EEPs regardless of the decision?

kikofernandez (approved):
`internal_export` is compliant with the EEP procedure and will be posted in the EEP index

kikofernandez:
We thank you for your contribution.

In this case, we have decided to reject EEP 67 on the basis that documentation attributes, introduced in OTP-27, is expressive enough to cover internal exports. Anything that has the documentation attribute `-doc false.` or `-doc hidden.` does not produce the documentation and can be understood as an internal function.

One can throw warnings using xref or using Dialyzer, by adding a new option that throws a warning when a hidden module is used externally.

TD5:
> One can throw warnings using xref or using Dialyzer, by adding a new option that throws a warning when a hidden module is used externally.

Good to see the problem being solved, even if it wasn't in precisely the way this EEP proposed it. I think Erlang will really benefit from having a way to do this.

kikofernandez:
Yes, we really wanted a way to express internal functions, and we think this is a good way to do it (not the only one 😃  )

kuna-invariant:
Thank you everyone for the review and time spent on this. 

But reasoning behind this rejection is abysmal and disappointing!

PR #59, opened 2024-02-12
EEP68 - JSON

michalmuskala:
This EEP proposes adding a `json` module to the standard library

[Rendered](https://github.com/michalmuskala/eep/blob/json-eep/eeps/eep-0068.md)

### DIFF
--- a/eeps/eep-0068.md
+++ b/eeps/eep-0068.md
@@ -1,0 +1,369 @@
+    Author: Michał Muskała <micmus(at)whatsapp(dot)com>
+    Status: Draft
+    Type: Standards Track
+    Created: 12-02-2024
+    Erlang-Version:
+    Post-History:
+****
+# EEP 68: JSON library
+----
+
+## Abstract
+
+This EEP proposes introducing a module `json` to the Erlang standard
+library with support for encoding and decoding [JSON][1] documents
+from and to Erlang data structures. The main reason is to cover
+a gap in the Erlang standard library with regards to such a vastly
+popular and widespread data format.
+
+## Rationale
+
+JSON is commonly in many different use-cases:
+* by web services as a lightweight and human-readable data interchange format;
+* as a configuration language in static files;
+* as data interchange format by developer tooling;
+* and more.
+
+There are many existing JSON libraries for Erlang and other BEAM languages,
+however adding such a support to standard library would offer unique benefits.
+Most notably being able to use it in situations where leveraging third-party
+libraries is complex or cumbersome -- such as stand-alone escripts or
+fundamental tooling like a build system, or inside OTP itself.
+
+There have been previous attempts to bring JSON support into OTP, most notably
+[EEP 18][EEP], which ultimately weren't adopted previously for various reasons.
+However, I believe the time is right to revisit this subject with a fresh
+take on an interface such support could take.
+
+JSON is a well defined format specified in parallel in [RFC 8259][RFC] and
+[ECMA 404][ECMA], however how this representation should be translated
+into Erlang is not fully clear since the data structures don't present
+a direct, 1:1 mapping. To help with this, this EEP proposes an interface
+that presents both a convenient and "cannonical" simple API, as well
+as an extensible and highly-customisable API with common underlying
+implementation.
+
+This EEP proposes a JSON library which:
+* should be easy to adopt in large codebases using one of the popular,
+  existing, open-source JSON libraries;
+* will allow the existing open-source libraries with custom features
+  (like support for Elixir protocols) to become thin wrappers around
+  this library;
+* will improve, or at least not regress, performance compared to
+  leading open-source JSON libraries.
+
+The proposed JSON library will provide:
+* JSON encoding, allowing for single-pass encoding of custom data types –-
+  in particular, for Elixir, integrating with a protocol through a thin layer
+  (implemented outside of OTP);
+* JSON decoding with some streaming support allowing to decode messages that
+  don't fully fit into memory;
+* JSON decoding with support for decoding values split across separate
+  messages without fully concatenating them upfront;
+* focus on high-performance encoding and decoding;
+* full conformance to [RFC 8259][RFC] and [ECMA 404][ECMA] standards,
+  the decoder should pass the entire [JSONTestSuite][JSONTestSuite];
+* simple API for common use-cases with canonical data type mapping.
+
+## Design choices
+
+### Data mapping
+
+We propose, in the "cannonical" API to map JSON data structues to
+Erlang and back in the following way:
+
+| **Decoding from JSON** | **Erlang**           | **Encoding into JSON** |
+|------------------------|----------------------|------------------------|
+| Number                 | integer() \| float() | Number                 |
+| Boolean                | true \| false        | Boolean                |
+| Null                   | null                 | Null                   |
+| String                 | binary()             | String                 |
+|                        | atom()               | String                 |
+| Array                  | list()               | Array                  |
+| Object                 | #{binary() => _}     | Object                 |
+|                        | #{atom() => _}       | Object                 |
+|                        | #{integer() => _}    | Object                 |
+
+Erlang has generally a richer value system than JSON, therefore
+there's generally more types that can be encoded into JSON,
+even if they can never be produced directly by the decoder.
+
+However, with the flexible API, as demonstrated below, the user will
+be able to customize the decoding & encoding routines to produce and
+consume any Erlang term as necessary in the particular application.
+
+### Streaming vs value-based parser
+
+When it comes to data-structure parsers it's common to encounter two
+types: ones that given the data produce a complete parsed value,
+and others the same data produce a stream of events that can later
+be processed to extract values.
+
+The first kind, which we'll call here value-based, is generally simpler,
+usually more efficient, and more convient to use. The second one offers
+unique advantages in specific use-cases: for example, where data
+can't fully fit into memory.
+
+For the proposed `json` library this EEP suggests a hybrid approach.
+
+First, a simple, value-based API:
+
+```erlang
+-type value() ::
+    integer() |
+    float() |
+    boolean() |
+    null |
+    binary() |
+    list(value()) |
+    #{binary() => value()}.
+
+-spec decode(binary()) -> value().
+```
+
+Error handling is achieved through exceptions. The following errors
+are possible:
+```erlang
+-type error() ::
+    unexpected_end |
+    {unexpected_sequence, binary()} |
+    {invalid_byte, byte()}
+```
+
+The exceptions might be enhanced through the [Error Info][ERRINFO] mechanism
+with additional meta-data like byte offset where the error occured.
+
+For the advanced and customizable API, this EEP proposes a callback-based
+API that the decoder will use to produce values from the data it parses.
+
+```erlang
+-type from_binary_fun() :: fun((binary()) -> dynamic()).
+-type array_start_fun() :: fun((Acc :: dynamic()) -> ArrayAcc :: dynamic()).
+-type array_push_fun() :: fun((Value :: dynamic(), Acc :: dynamic()) -> NewAcc :: dynamic()).
+-type array_finish_fun() :: fun((ArrayAcc :: dynamic()) -> dynamic()).
+-type object_start_fun() :: fun((Acc :: dynamic()) -> ObjectAcc :: dynamic()).
+-type object_push_fun() :: fun((Key :: dynamic(), Value :: dynamic(), Acc :: dynamic()) -> NewAcc :: dynamic()).
+-type object_finish_fun() :: fun((ObjectAcc :: dynamic()) -> dynamic()).
+
+-type decoders() :: #{
+    empty_array => term(),
+    array_start => array_start_fun(),
+    array_push => array_push_fun(),
+    array_finish => array_finish_fun(),
+    empty_object => term(),
+    object_start => object_start_fun(),
+    object_push => object_push_fun(),
+    object_finish => object_finish_fun(),
+    float => from_binary_fun(),
+    integer => from_binary_fun(),
+    string => from_binary_fun(),
+    null => term()
+}.
+
+-spec decode(binary(), Acc :: dynamic(), decoders()) ->
+    {Value :: dynamic(), FinalAcc :: dynamic(), Rest :: binary()}.
+```
+
+This allows the user to fully customize the decoded format, including
+features seen in open-source JSON libraries:
+* decoding string keys as atoms;
+* decoding objects as lists of pairs;
+* decoding floats as custom structures with decimal precision;
+* decoding `null` as another atom, in particular `undefined` or `nil`;
+* using `binary:copy/1` on strings that will be retained in memory;
+* decoding multiple JSON messages from a single binary blob;
+* and more.
+
+Furthermore, this allows the user to only retain parts of the data structure
+to achieve results similar to using a streaming SAX-like parser for data
+that does't fully fit into memory.
+
+All the callbacks are optional and have a default value correspnding to the
+"simple" API behaviour and using lists as accumulators.
+
+### Incomplete data parsing
+
+We propose a future enhancement to the full `decode/3` API, where
+it can return an `{incomplete, continuation()}` value that can be used to
+decode values split across multiple binary blobs (for example as received
+from a TCP socket).
+
+```erlang
+-spec decode_continue(binary(), continuation()) ->
+    {Value :: dynamic(), FinalAcc :: dynamic(), Rest :: binary()} |
+    {incomplete, continuation()}.
+```
+
+### Encoding API
+
+For encoding this EEP again proposes two separate sets of APIs.
+A simple API using "cannonical" data types:
+
+```erlang
+-type encode_value() ::
+    integer() |
+    float() |
+    boolean() |
+    null |
+    binary() |
+    atom() |
+    list(encode_value()) |
+    #{binary() | atom() | integer() => encode_value()}.
+
+-spec encode(encode_value()) -> iodata().
+```
+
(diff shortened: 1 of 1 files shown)


eeps/eep-0068.md
```diff
@@ -0,0 +1,369 @@
...
+There have been previous attempts to bring JSON support into OTP, most notably
+[EEP 18][EEP], which ultimately weren't adopted previously for various reasons.
+However, I believe the time is right to revisit this subject with a fresh
+take on an interface such support could take.
+
+JSON is a well defined format specified in parallel in [RFC 8259][RFC] and
+[ECMA 404][ECMA], however how this representation should be translated
+into Erlang is not fully clear since the data structures don't present
+a direct, 1:1 mapping. To help with this, this EEP proposes an interface
+that presents both a convenient and "cannonical" simple API, as well
```
josevalim, line 42:
```suggestion
that presents both a convenient and "canonical" simple API, as well
```

eeps/eep-0068.md
```diff
@@ -0,0 +1,369 @@
...
+### Example of a custom encoder
+
+An example of a custom encoder that would support using a heuristic
+to differentiate pais of object-like key-value lists from plain
+lists of values could look as follows:
+```erlang
+custom_encode(Value) -> json:encode(Value, fun encoder/2).
+
+encoder(null, _Encode) -> <<"\"null\"">>;
+encoder(nil, _Encode) -> <<"null">>;
```
josevalim, line 333: Are these two clauses meant to be removed?

```suggestion
```

eeps/eep-0068.md
```diff
@@ -0,0 +1,369 @@
...
+-type object_start_fun() :: fun((Acc :: dynamic()) -> ObjectAcc :: dynamic()).
+-type object_push_fun() :: fun((Key :: dynamic(), Value :: dynamic(), Acc :: dynamic()) -> NewAcc :: dynamic()).
+-type object_finish_fun() :: fun((ObjectAcc :: dynamic()) -> dynamic()).
+
+-type decoders() :: #{
+    empty_array => term(),
+    array_start => array_start_fun(),
+    array_push => array_push_fun(),
+    array_finish => array_finish_fun(),
+    empty_object => term(),
```
josevalim, line 153: Is there a reason for having specific callbacks for empty_array and empty_object?

eeps/eep-0068.md
```diff
@@ -0,0 +1,369 @@
...
+-spec format(iodata()) -> iodata().
+-spec format(iodata(), format_option()) -> iodata().
+```
+
+## Reference Implementation
+
+[PR-8111][PR] Implements the `encode/1`, `encode/2`, `decode/1`, and `decode/3`
+functions as proposed in this EEP.
+The formatting API and the support for incomplete message decoding is left
+as a follow-up taskk.
```
josevalim, line 278:
```suggestion
as a follow-up task.
```

eeps/eep-0068.md
```diff
@@ -0,0 +1,369 @@
...
+
+This EEP further proposes an additional API for formatting (and pretty-printing)
+JSON messages. This API consists of transforming a textual JSON message into
+a formatted JSON message.
+This is the most flexible solution that orthogonally supports
+formatting results of custom encoding functions like described above,
+without adding the burden of complex formatting options in the middle of the
+encoders.
+Formatting isn't usually done in critical hot-paths of high-performance
+services, thgerefore the overhead of a two-pass formatting is deemed acceptable.
```
josevalim, line 261:
```suggestion
services, therefore the overhead of a two-pass formatting is deemed acceptable.
```

eeps/eep-0068.md
```diff
@@ -0,0 +1,369 @@
...
+```erlang
+-type from_binary_fun() :: fun((binary()) -> dynamic()).
+-type array_start_fun() :: fun((Acc :: dynamic()) -> ArrayAcc :: dynamic()).
+-type array_push_fun() :: fun((Value :: dynamic(), Acc :: dynamic()) -> NewAcc :: dynamic()).
+-type array_finish_fun() :: fun((ArrayAcc :: dynamic()) -> dynamic()).
+-type object_start_fun() :: fun((Acc :: dynamic()) -> ObjectAcc :: dynamic()).
+-type object_push_fun() :: fun((Key :: dynamic(), Value :: dynamic(), Acc :: dynamic()) -> NewAcc :: dynamic()).
+-type object_finish_fun() :: fun((ObjectAcc :: dynamic()) -> dynamic()).
+
+-type decoders() :: #{
```
josevalim, line 148: Are you concerned about the cost of having several callbacks? API wise and performance wise, what do you think about a single array and object callback, that returns both push and finish functions?

```
-type object_fun(Acc :: dynamic()) -> {ArrayAcc, object_push_fun(), object_finish_fun()}.
```

eeps/eep-0068.md
```diff
@@ -0,0 +1,369 @@
...
+| String                 | binary()             | String                 |
+|                        | atom()               | String                 |
+| Array                  | list()               | Array                  |
+| Object                 | #{binary() => _}     | Object                 |
+|                        | #{atom() => _}       | Object                 |
+|                        | #{integer() => _}    | Object                 |
+
+Erlang has generally a richer value system than JSON, therefore
+there's generally more types that can be encoded into JSON,
+even if they can never be produced directly by the decoder.
```
essen, line 89: Yes but this also means that `Value = decode(encode(Value))` is not always true. Even if you customise decoding, this may not be true. If you never use atoms (and floats?), then `Value` should remain the same before and after. This should at the very least be mentioned as a caveat.

michalmuskala: Yes, that's correct.
It's worth pointing out that this property doesn't hold in almost all the JSON libraries in dynamic languages I can think of - not in JS, not in Ruby, not in Python, and neither in any of the existing Erlang or Elixir libraries.
That said, I agree that mentioning this makes sense

eeps/eep-0068.md
michalmuskala: The implementation for decoding empty objects and arrays was quite awkward with calling `Decode#array_finish(Decode#array_start(Acc))`. It's not a hard requirement, but it made the implementation slightly more "elegant".

eeps/eep-0068.md
```diff
@@ -0,0 +1,372 @@
...
+  object_push(Str4, Arr2, Acc11) => Acc16
+object_finish(Acc16) => Obj3
+% final decode/3 return
+{Obj3, Acc16, <<"">>}
+```
+
+### Example of a custom encoder
+
+An example of a custom encoder that would support using a heuristic
+to differentiate pais of object-like key-value lists from plain
```
dgud, line 332: pais => pairs

eeps/eep-0068.md
```diff
@@ -0,0 +1,372 @@
...
+
+### Streaming vs value-based parser
+
+When it comes to data-structure parsers it's common to encounter two
+types: ones that given the data produce a complete parsed value,
+and others the same data produce a stream of events that can later
+be processed to extract values.
+
+The first kind, which we'll call here value-based, is generally simpler,
+usually more efficient, and more convient to use. The second one offers
```
legoscia, line 108:
```suggestion
usually more efficient, and more convenient to use. The second one offers
```

eeps/eep-0068.md
```diff
@@ -0,0 +1,372 @@
...
+* decoding objects as lists of pairs;
+* decoding floats as custom structures with decimal precision;
+* decoding `null` as another atom, in particular `undefined` or `nil`;
+* using `binary:copy/1` on strings that will be retained in memory;
+* decoding multiple JSON messages from a single binary blob;
+* and more.
+
+Furthermore, this allows the user to only retain parts of the data structure
+to achieve results similar to using a streaming SAX-like parser for data
+that does't fully fit into memory.
```
legoscia, line 184:
```suggestion
that doesn't fully fit into memory.
```

eeps/eep-0068.md
```diff
@@ -0,0 +1,372 @@
...
+
+All the callbacks are optional and have a default value correspnding to the
```
legoscia, line 186:
```suggestion
All the callbacks are optional and have a default value corresponding to the
```

eeps/eep-0068.md
```diff
@@ -0,0 +1,372 @@
...
+### Data mapping
+
+We propose, in the "canonical" API to map JSON data structues to
+Erlang and back in the following way:
+
+| **Decoding from JSON** | **Erlang**           | **Encoding into JSON** |
+|------------------------|----------------------|------------------------|
+| Number                 | integer() \| float() | Number                 |
+| Boolean                | true \| false        | Boolean                |
+| Null                   | null                 | Null                   |
```
manuel-rubio, line 79: only asking, why not `undefined`? I think that `undefined` is usually accepted as the Erlang null value.

eeps/eep-0068.md
```diff
@@ -0,0 +1,372 @@
...
+There are many existing JSON libraries for Erlang and other BEAM languages,
+however adding such a support to standard library would offer unique benefits.
+Most notably being able to use it in situations where leveraging third-party
+libraries is complex or cumbersome -- such as stand-alone escripts or
+fundamental tooling like a build system, or inside OTP itself.
+
+There have been previous attempts to bring JSON support into OTP, most notably
+[EEP 18][EEP], which ultimately weren't adopted previously for various reasons.
+However, I believe the time is right to revisit this subject with a fresh
+take on an interface such support could take.
```
zuiderkwast, line 36: From Ericsson's point of view, I believe the key difference as to why JSON can be acceptable in OTP now and not in the past, is the introduction of JSON in the 5G network standards.

As opposed to earlier mobile networks (2G, 3G, 4G), the 5G network uses JSON REST-APIs over HTTP/2 in what is called "Service Based Architecture". The specifications are open but it's quite difficult to find the right ones. 3GPP TS 29.500 is a good starting point for this though: https://www.etsi.org/deliver/etsi_ts/129500_129599/129500/16.04.00_60/ts_129500v160400p.pdf

The the older 2G, 3G and 4G networks there was no use of JSON or TCP or HTTP or almost any web protocols. There were telecom-specific protocols (such as DIAMETER) for everything.

dgud (commented):
I'm wondering if we should NOT stack the accumulator for the user, and let the user handle that and also make the `array_finish/1` and `object_finish/1` return {Obj, NewAcc}.

As it is implemented it is not useful if I for example during parsing also want to count the 
numbers of sharks in my array of aquariums.

I think that the Acc0 sent in is the one returned from decode/3 shows that the current proposal is strange.

I think I agree with @josevalim empty_array and empty_object, 
from a user perspective it seems strange to handle empty objects separately, though
looking at the code I can understand why.

michalmuskala:
@dgud yes, you're correct. I think my immediate concern was the overhead of the extra tuples each time we finalise a value - it does really add up. Let me try some benchmarks to see how much of an overhead that would be

michalmuskala:
I pushed an update that removes the `_empty` callbacks and updates the `_finish` callbacks to also receive the "old accumulator" (as was passed to `_start`), and expects them to return a tuple with the value and next accumulator - this allows threading a value in the accumulator throughout all the calls.
With some other optimisations the performance of the new encoder is still faster than other pure-Erlang implementations so I think it's worth it.

garazdawi:
When designing the API for [logger filters](https://www.erlang.org/doc/man/logger.html#type-filter) we decided to not use a `function()`, instead we use `{function(), term()}` which makes it possible to use export funs as filters. The reason for this was two fold:

1. It allows the code for the module the filter is in to be upgraded without having to re-configure logger with the new fun.
2. It makes it possible to configure filters in `sys.config` files.

Do you think the same reasons apply to json? If so, should the encoder/decoder types be `{function(),term()}` instead of the current `function()`? Or should we also leave that as a possible extension?

The reason I started thinking about this is because I had a look at Jason and it supports a couple of different ways to escape strings (javascript, html, etc). The way to implement the same thing with EEP-68 json would be to implement an encoder callback for binary, but you would need to have different funs for each encoding type as you cannot pass any arguments to the encoder.

eeps/eep-0068.md
```diff
@@ -0,0 +1,387 @@
...
+Error handling is achieved through exceptions. The following errors
+are possible:
+
+    -type error() ::
+        unexpected_end |
+        {unexpected_sequence, binary()} |
+        {invalid_byte, byte()}
+
+The exceptions might be enhanced through the [Error Info][ERRINFO] mechanism
+with additional meta-data like byte offset where the error occured.
```
garazdawi, line 144:
```suggestion
with additional meta-data like byte offset where the error occurred.
```

michalmuskala:
@garazdawi How are the functions called then? Is is `Fun(<current_args>, Term)` or is it something like `apply(Fun, CurrentArgs ++ Term)`?
For the first case, we could consider it, however it would make the "simple case" more complex.
The reason why I'm hesitant about the second case is that with fixed number of arguments we can convince compiler in `parse_decoder` that all callbacks have the right number of arguments and thus emit the more efficient fun calls that don't check arguments inside of the main decoding loops.

As to the various escape functions in Jason - they're purpose was actually questioned to me, and I didn't have a clear answer, other than replicating the API of the older library Poison.

In particular the `javascript` case is no longer necessary as of 2019 since Javascript got explicitly re-defined as a subset of JSON in https://tc39.es/proposal-json-superset/
And for the `html` case it's actually not fully clear what it's supposed to protect against as the escaping seems to be not-enough and too-much at the same time - it was also broken for a really long time in Jason without any complaints (see https://github.com/michalmuskala/jason/issues/109). 
That's why this proposed implementation doesn't include either of those modes.

garazdawi:
> @garazdawi How are the functions called then? Is is `Fun(<current_args>, Term)` or is it something like `apply(Fun, CurrentArgs ++ Term)`? For the first case, we could consider it, however it would make the "simple case" more complex. The reason why I'm hesitant about the second case is that with fixed number of arguments we can convince compiler in `parse_decoder` that all callbacks have the right number of arguments and thus emit the more efficient fun calls that don't check arguments inside of the main decoding loops.

They are called as the first example, I.e. `fun(CurrentArgs, Term)`. But I think we should leave it as is, logger has an enough different use case to json that there is no reason to make things more complicated than they need to be.

I only mentioned the string escaping as an example of what you could want to do with the custom encoder interface. I don’t see any reason to add it to this module.

eeps/eep-0068.md
```diff
@@ -0,0 +1,387 @@
...
+However, with the flexible API, as demonstrated below, the user will
+be able to customize the decoding & encoding routines to produce and
+consume any Erlang term as necessary in the particular application.
+
+**Note**: A decode-encode rountrip might not produce the same data,
+even with custom decoders -- since JSON has such a limited data-type
+options, compared to Erlang, some information will be commonly be lost,
+for example, coercing all keys in maps to binaries.
+
+Streaming vs value-based parser
```
TD5, line 107: > Streaming vs value-based parser

"Streaming vs. Batch" is perhaps the more standard terminology?

michalmuskala: This isn't really a batch parser - it still parses single value at a time. To me "batch" implies processing a batch of objects at a time. 
The notation I've seen used was "streaming vs DOM", but I found "value" a good replacement to "DOM" that doesn't need lengthy explanations

saleyn:
Would the suggested implementation for the JSON encoding/decoding be NIF-based on pure Erlang?  What is the benefit of having this functionality to be a part of the Erlang distribution?  There are already multiple good open-source JSON libraries available, and reinventing the wheel might not be worth the effort?  Being an author of [one of such libraries](https://github.com/saleyn/simdjsone), I might be biased. :)

okeuday:
@saleyn The code should be more trustworthy if it is done in Erlang and not NIF-based, when considering the code initially added and its potential changes in the future.  Erlang/OTP already provides `xmerl` for XML parsing and JSON has become more popular, so it makes sense to add JSON support in Erlang/OTP as a central place where the API can help people avoid mistakes like [CVE-2017-12635](https://nvd.nist.gov/vuln/detail/CVE-2017-12635).

michalmuskala:
The proposed API for both encoding and decoding includes callbacks - since it's not possible to call back to Erlang from NIFs, this can't be implemented as a pure NIF. 

There's a pure Erlang implementation PR opened against the OTP https://github.com/erlang/otp/pull/8111. 

While the open-source implementations are great - this proposed one has even better performance and a significantly more flexible interface. It outperfors all pure Erlang implementations that I know of, and most of time time even the NIF-based jiffy (I failed to compile the [simdjsone](https://github.com/saleyn/simdjsone/issues/6) project). 

Having it in standard distribution also makes it possible to use JSON in simple escripts (greatly enhancing the utility of Erlang as a scripting language) and makes it easy to use JSON in fundamental tooling like build systems (e.g. producing metadata about project layout for other tools like language servers), or in other parts of OTP itslef.

okeuday:
@michalmuskala My memory of testing jiffy in the past was that it became slower when using the dirty thread pool, though it had the possibility of blocking VM scheduler threads if the JSON being parsed was too large (when not using the dirty thread pool).  Avoiding performance and reliability problems like that with a pure Erlang implementation will be best.

nixxquality:
I would like to submit this for your consideration: https://erlangforums.com/t/jsonpull-a-json-pull-parser-for-erlang/3306
If you have time to watch the original talk, please do, but otherwise I've already summed up my thoughts in the post above.

While this can lead to verbose code, it opens up a system where anything can be built on top, whether that is simple functions like my construct_list or even different kinds of parsers, like the ones already implemented here. Additionally, it could lead to third party libraries built on a standard one that immediately parses into expected types.

saleyn:
> It outperforms all pure Erlang implementations that I know of, and most of the time even the NIF-based jiffy (I failed to compile the [simdjsone](https://github.com/saleyn/simdjsone/issues/6) project)

Could you post an issue there with more details if you had compilation issues?  I have that library working in several deployments without issues.

PR #60, opened 2024-04-11
Added eep draft for nominal types

lucioleKi:
This commit adds the eep draft for nominal types. I plan to make a post on Erlang forum, introducing nominal types. The post will refer to this pull request.

### DIFF
--- a/eeps/eep-XXX.md
+++ b/eeps/eep-XXX.md
@@ -159,7 +159,7 @@ nominal types in Dialyzer.
 New Features
 ------
 
-- Now, doing pattern-matching on opaques will raise a warning without 
+- Now, doing pattern-matching on opaques will raise a warning without
 stopping Dialyzer's analysis. (Previously, doing pattern-matching on opaques
 raised an error and caused Dialyzer to stop the analysis, and so Dialyzer
 could not catch other errors.)
@@ -181,13 +181,15 @@ for opaques as for nominals.
 
 Reference Implementation
 ========================
+
 Current implementation: <https://github.com/lucioleKi/otp/tree/cleanup>
 
 Backward compatibility
 ========================
+
 Code that contains no opaque type or does not use Dialyzer has no  change.
 
-Code that contains opaque type(s) and uses Dialyzer may experience changes 
+Code that contains opaque type(s) and uses Dialyzer may experience changes
 mentioned above.
 
 [1]: https://flow.org/en/docs/lang/nominal-structural/#in-flow
--- a/eeps/eep-XXX.md
+++ b/eeps/eep-XXX.md
@@ -153,11 +153,6 @@ in `s()`.
           -nominal s() :: nominal_1().
           -nominal nominal_1() :: nominal_2().
           -nominal nominal_2() :: t().
-- One case where `s()` is *not* a nominal subtype of `t()`:
-    - `s()` is defined as a union type of `t()` and some other type. The
-    nesting cannot contain other types before reaching `t()`.
-
-          -nominal s() :: t() | 'foo'.
 
 A non-nominal type is **compatible** with a nominal or non-nominal type
 if they share common values. Empirically, the function `erl_types:t_inf/2`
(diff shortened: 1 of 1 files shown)


RaimoNiskanen:
I am missing a more formal definition for the suggested nominal types.
As it stands now I have a hard time to follow the reasoning and examples...

Such as: 

A function that has a `-spec` that states an argument to be nominal type `a/0`, accepts:
* Nominal type `a/0`
* A supertype to nominal type `a/0`
* A compatible structural type.
* ¿ A subtype to nominal type `a/0` ?

A function that has a `-spec` that it returns a nominal type `b/0` may return:
* Nominal type `b/0`
* A supertype to nominal type `b/0`
* A compatible structural type.
* ¿ A subtype to nominal type `a/0` ?

A function that has a `-spec` that states an argument to be structural type `c/0`, accepts:
* A compatible structural type.
* A nominal subtype of structural type `c/0`

A function that has a `-spec` that it returns a structural type `d/0` may return:
* A compatible structural type.
* A nominal subtype of structural type `d/0`

"supertype" and "subtype" also needs to be defined, and maybe "compatible structural type"

zuiderkwast:
That's what I though *opaque* types are. My interpretation of opaque types is that you can't assume anything about what they are so you need to treat them as nominal types.

Type checkers probably differ here though. I don't know what Dialyzer does.

lucioleKi:
@RaimoNiskanen It seems that my intentional avoidance of subtyping in the EEP draft causes more problems than benefits. I used 'nested nominals' instead of 'subtyping', because I wanted to emphasize that the supertypes and subtypes accepted are **nominal supertypes** and **nominal subtypes**. In Erlang, the only way to declare a nominal subtype is to declare a nested nominal of the form `-nominal a() :: b()` where `b()` is another nominal type. 

I'd phrase it as follows (same for `b/0`):

A function that has a -spec that states an argument to be nominal type `a/0`, accepts:

- Nominal type `a/0`

- A nominal supertype to nominal type `a/0`

- A compatible structural type.

- A nominal subtype to nominal type `a/0`

RaimoNiskanen:
@lucioleKi, and what gives for return types?

lucioleKi:
@zuiderkwast Yes, opaque types have always been type-checked by names in Dialyzer, just with more restrictions. Dialyzer assumes no knowledge about opaque's structure, does not consider subtyping, etc. That's why the proposed nominal type-checking can cover opaque type-checking internally in Dialyzer. Nominal types are 'transparent' though. This EEP suggests a way to support nominal types and type-checking without opacity.

lucioleKi:
@RaimoNiskanen The examples `c/0` and `d/0` behaves as you described. For a nominal type `a() :: T` and a structural type `S`, `S` is a 'compatible structural type' of `a()` when the intersection of `T` and `S` is non-empty. (I'm not sure if this definition is too technical. This is how Dialyzer does it internally.)  

A function that has a -spec that it returns a nominal type b/0 may return the following without Dialyzer raising an error:

- Nominal type `b/0`

- A nominal supertype to nominal type `b/0`

- A compatible structural type.

- A nominal subtype to nominal type `b/0`

RaimoNiskanen:
Great!  I think these definitions should be in the EEP.

eeps/eep-XXX.md
```diff
@@ -0,0 +1,278 @@
...
+A non-nominal type is **compatible** with a nominal or non-nominal type
+if they share common values. Empirically, the function `erl_types:t_inf/2`
+can compute the intersection of 2 types. Two types that have a non-empty
+intersection are structurally compatible.
+
+- Examples:
+    - 4711 and 42 are not structurally compatible.
+    - 4711 and `integer()` are structurally compatible. (Their intersection
+    is the value 4711.)
+    - `list(any_type)` and `[]` are structurally compatible.
```
kikofernandez, line 171: Since you write the reason, may it helps the reader if you write their intersection as in the prev. point above

kikofernandez (commented):
I think this was a good addition. I have just one minor comment

kikofernandez (approved):

PR #64, opened 2024-10-09
Update record names as reserved words/variables and new record definition syntax

choptastic:
Hello,

This is in reference to [PR-7873](https://github.com/erlang/otp/pull/7873) and is made on @RaimoNiskanen's [recommendation](https://github.com/erlang/otp/pull/7873#issuecomment-2346314677).

@RaimoNiskanen, I did not include you as an author on the EEP, but since you implemented the new record syntax, you obviously deserve to be on it.  I just didn't want to include you without your blessing first.

### DIFF
--- a/eeps/eep-0072.md
+++ b/eeps/eep-0072.md
@@ -230,8 +230,8 @@ defining records. Since usage of the syntax for getting, setting, matching, etc
 (e.g. `#rec{a=x,y=b}`) occurs far more commonly than defining, it only feels
 natural that the definition syntax would mirror usage.
 
-Further, the syntax in Erlang's type system to define records also matches the
-newly proposed define syntax.
+For more symmetry, the syntax in Erlang's type system to define records also
+matches the newly proposed define syntax.
 
 Thus, I feel that sharing the existing usage and type syntax with the
 definition system would likely become the default/preferred way, and that the
--- a/eeps/eep-0072.md
+++ b/eeps/eep-0072.md
@@ -10,8 +10,8 @@ EEP 72: Reserved words and Variables as record names, and enhancement to definit
 Abstract
 ========
 
-This EEP loosens some of the restrictions around record names such no longer is
-it necessary to quote record names when those records are named with reserved
+This EEP loosens some of the restrictions around record names to make it
+no longer necessary to quote them when they are named with reserved
 words (`#if` vs `#'if'`) or words with capitalized first characters (`#Hello`
 vs `#'Hello'`).
 
--- a/eeps/eep-0072.md
+++ b/eeps/eep-0072.md
@@ -21,15 +21,15 @@ following record definitions would all be valid and identical:
 
 ```erlang
 -record('div', {a :: integer(), b :: integer()}).
--record #'div'{a :: integer(), b :: integer()}.
 -record #div{a :: integer(), b :: integer()}.
 ```
 
-The latter two are proposed new syntax.  The following would also be valid
+The latter one is proposed new syntax.  The following would also be valid
 and identical since parentheses are optional in attributes:
 
 ```erlang
 -record 'div', {a :: integer(), b :: integer()}.
+-record #'div'{a :: integer(), b :: integer()}.
 -record(#'div'{a :: integer(), b :: integer()}).
 -record(#div{a :: integer(), b :: integer()}).
 ```
--- a/eeps/eep-0072.md
+++ b/eeps/eep-0072.md
@@ -43,6 +43,7 @@ names to be consistent with the rest of the language's use of atoms.
 All atoms in Erlang can be denoted with single quotes. Some examples:
 
 For example:
+
 ```erlang
 'foo'.
 'FOO'.
--- a/eeps/eep-0072.md
+++ b/eeps/eep-0072.md
@@ -12,8 +12,8 @@ Abstract
 
 This EEP loosens some of the restrictions around record names to make it
 no longer necessary to quote them when they are named with reserved
-words (`#if` vs `#'if'`) or words with capitalized first characters (`#Hello`
-vs `#'Hello'`).
+words (`#if` vs `#'if'`) or words with capitalized first characters (terms that
+currently would be treated as variables, for example `#Hello` vs `#'Hello'`).
 
 This EEP also proposes to add a new record-like syntax to the record
 definitions (also adopting the above syntactical changes), so that the
--- a/eeps/eep-0072.md
+++ b/eeps/eep-0072.md
@@ -196,6 +196,7 @@ the new syntax if your code uses the new syntax rules.
 
 Broader Concerns and Points of Discussion
 =========================================
+
 While the new definition syntax creates some degree of symmetry around record
 usage, there do remain some concerns and inconsistencies:
 
--- a/eeps/eep-0072.md
+++ b/eeps/eep-0072.md
@@ -91,7 +91,7 @@ go() ->
 ```
 
 While this approach is consistent with atom usage in the language, for reserved
-words, this makes the record syntax *feel* inconsistent if you have a need for
+words and capitalized atoms, this makes the record syntax *feel* inconsistent if you have a need for
 naming a record with a reserved word (or term with a capital first letter). In
 this case, it almost guarantees a user won't use a record named 'if',
 'receive', 'fun', etc even though there may very well be a valid use case for
@@ -183,7 +183,7 @@ Implementation
 
 To update the syntax for using records, we can safely augment the parser to
 change its already existing record handling of `'#' atom '{' ... '}'` and `'#' atom '.' atom` into `'#' record_name '{' ... '}'` and `'#' record_name '.' atom`, and define
-`record_name` to be `atom`, `var`, or `reserver_word`.
+`record_name` to be `atom`, `var`, or `reserved_word`.
 
 To update the record definition syntax, we can simply add a few new
 modifications to the `attribute` Nonterminal to allow `'#' record_name` as name for the `record` attribute, instead of `atom` as for generic attributes.
@@ -222,6 +222,7 @@ should try to hide it. These are remaining concerns and inconsistencies:
    feel that the symmetry between the usage and the definition would likely
    become the default/preferred way, and that the original syntax remain for
    backwards compatibility.
+3. Another 
 
 Reference Implementation
 ========================
--- a/eeps/eep-0072.md
+++ b/eeps/eep-0072.md
@@ -17,7 +17,7 @@ vs `#'Hello'`).
 
 This EEP also proposes to add a new record-like syntax to the record
 definitions (also adopting the above syntactical changes), so that the
-following record definitions would all be valid and identical:
+following record definitions would be valid and identical:
 
 ```erlang
 -record('div', {a :: integer(), b :: integer()}).
--- a/eeps/eep-0072.md
+++ b/eeps/eep-0072.md
@@ -1,0 +1,147 @@
+    Author: Jesse Gumm <ja(at)gumm(dot)io>
+    Status: Draft
+    Type: Standards Track
+    Created: 08-Oct-2024
+    Post-History:
+****
+EEP 72: Reserved words and Variables as record names, and enhancement to definition syntax
+----
+
+Abstract
+========
+
+This EEP loosens some of the restrictions around record names such no longer is
+it necessary to quote record names when those records are named with reserved
+words (`#if` vs `#'if'`) or words with capitalized first characters (`#Hello`
+vs `#'Hello'`).
+
+```erlang
+#'if'{
+
+This EEP also proposes to add a new record-like syntax to the record
+definitions (also adopting the above syntactical changes), such that the
+following record definitions would be identical:
+
+```erlang
+-record('div', {a :: integer(), b :: integer()}).
+-record #'div'{a :: integer(), b :: integer()}.
+-record #div{a :: integer(), b :: integer()}.
+```
+
+Motivation
+==========
+
+Record names are atoms. As such, the current Erlang syntax requires the record
+names to be consistent with the rest of the language's use of atoms.
+
+All atoms in Erlang can be denoted with single-quotes. Some examples:
+
+For example:
+```erlang
+'foo'.
+'FOO'.
+'foo-bar`.
+```
+
+But, conveniently, simple atoms (all alphanumeric, underscores (`_`) or
+at-symbols (`@`) with the first character being a lower-case letter and not one
+of the 20+ reserved words), in all contexts can be invoked without the
+necessary wrapping quotes. Some examples:
+
+```erlang
+foo.
+
+```
+
+Conveniently, this also means that records named with simple atoms can be
+invoked and used without having to quote the atoms. For example:
+
+```erlang
+-record(foo, {a, b}).
+
+go() ->
+    X = #foo{a = 1, b = 2},
+    Y = X#foo{a = something_else},
+    Z = Y#foo.a,
+    ...
+```
+
+Unfortunately, that also means that records named with anything that doesn't
+fit the "simple atom" pattern must be wrapped in quotes in definition and
+usage. For example:
+
+```erlang
+-record('div', {a, b}).
+
+go() ->
+    X = #'div'{a = 1, b = 2},
+    Y = X#'div'{a = something_else},
+    Z = Y#'div'.a,
+    ...
+```
+
+While this approach is consistent with atom usage in the language, for reserved
+words, this makes the record syntax *feel* inconsistent if you have a need for
+naming a record with a reserved word (or term with a capital first letter).  In
+this case, it almost guarantees a user won't use a record named 'if',
+'recieve', 'fun', etc even though there may very well be a valid use-case for
+such a name.  The most common use-case that comes to mind from the Nitrogen Web
+Framework.  Since HTML has a `div` tag, Nitrogen (which represents HTML tags
+using Erlang records) should naturally have a `#div` record, however, due to
+'div' being a reserved word, the record `#panel` is used instead to save the
+programmer from having to invoke `#'div'`, which feels unnatural and awkward.
+
+Specification
+=============
+
+This EEP simplifies the above example by
+
+1. Allowing reserved words and variables to be used without quotes for record
+   names, and
+2. Simplifying the definition such that the syntax between record definition
+   and record usage becomes more consistent.
+
+With the changes from this EEP, the above code becomes:
+
+```erlang
+-record #div{a, b}.
+
+go() ->
+    X = #div{a = 1, b = 2},
+    Y = X#div{a = something_else},
+    Z = Y#div.a,
+    ...
+```
+
+Implementation
+==============
+
+To update the syntax for using records, we can safely augment the parser to
+change `# atom` to `# record_name` and define `record_name` to be either
+`atom`, `var`, or `reserved_word`.
+
+To update the record definition syntax, we can simply add a few new
+modifications to the `attribute` Nonterminal.
+
+Backwards Compatibility
+=======================
+
+As this EEP only adds new syntax, the vast majority existing codebases will
+still work, with the possible exception of AST/code analysis tools that are
+analyzing code using the new syntax.
+
+Syntax highlighting and code-completion tools may need to be updated to support
+the new syntax if your code uses the new syntax rules.
+
+Reference Implementation
+========================
+
+The reference implementation is provided in a form of pull-request on GitHub
+
+    https://github.com/erlang/otp/pull/7873
+
+Copyright
+=========
+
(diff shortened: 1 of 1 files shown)


eeps/eep-0072.md
```diff
@@ -0,0 +1,145 @@
...
+    Y = X#'div'{a = something_else},
+    Z = Y#'div'.a,
+    ...
+```
+
+While this approach is consistent with atom usage in the language, for reserved
+words, this makes the record syntax *feel* inconsistent if you have a need for
+naming a record with a reserved word (or term with a capital first letter).  In
+this case, it almost guarantees a user won't use a record named 'if',
+'recieve', 'fun', etc even though there may very well be a valid use-case for
```
josevalim, line 85:
```suggestion
'receive', 'fun', etc even though there may very well be a valid use-case for
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,145 @@
...
+    Post-History:
+****
+EEP 72: Reserved words and Variables as record names, and enhancement to definition syntax
+----
+
+Abstract
+========
+
+This EEP loosens some of the restrictions around record names such no longer is
+it necessary to quote record names when those records are named with reserved
```
RaimoNiskanen, line 14:
```suggestion
This EEP loosens some of the restrictions around record names to make it
no longer necessary to quote them when they are named with reserved
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,145 @@
...
+words (`#if` vs `#'if'`) or words with capitalized first characters (`#Hello`
+vs `#'Hello'`).
```
RaimoNiskanen, line 16:
```suggestion
words (`#if` vs `#'if'`) or words with capitalized first characters (variable
names, for example `#Hello`￼vs. `#'Hello'`).
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,145 @@
...
+
+This EEP also proposes to add a new record-like syntax to the record
+definitions (also adopting the above syntactical changes), such that the
```
RaimoNiskanen, line 19:
```suggestion
definitions (also adopting the above syntactical changes), so that the
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,145 @@
...
+following record definitions would all be valid and identical:
+
+```erlang
+-record('div', {a :: integer(), b :: integer()}).
+-record #'div'{a :: integer(), b :: integer()}.
+-record #div{a :: integer(), b :: integer()}.
+```
+
```
RaimoNiskanen, line 27:
````suggestion
The latter two are proposed new syntax.  The following would also be valid
and identical since parentheses are optional in attributes:

```erlang
-record 'div', {a :: integer(), b :: integer()}.
-record(#'div'{a :: integer(), b :: integer()}).
-record(#div{a :: integer(), b :: integer()}).
```
````

File: eeps/eep-0072.md [hunk 6]
```diff
@@ -0,0 +1,145 @@
...
+While this approach is consistent with atom usage in the language, for reserved
+words, this makes the record syntax *feel* inconsistent if you have a need for
+naming a record with a reserved word (or term with a capital first letter).  In
+this case, it almost guarantees a user won't use a record named 'if',
+'recieve', 'fun', etc even though there may very well be a valid use-case for
+such a name.  The most common use-case that comes to mind from the Nitrogen Web
+Framework.  Since HTML has a `div` tag, Nitrogen (which represents HTML tags
+using Erlang records) should naturally have a `#div` record, however, due to
+'div' being a reserved word, the record `#panel` is used instead to save the
+programmer from having to invoke `#'div'`, which feels unnatural and awkward.
```
RaimoNiskanen, line 90:
```suggestion
'div' being a reserved word (the integer division operator), the record `#panel`
is used instead to save the programmer from having to invoke `#'div'`,
which feels unnatural and awkward.
```

RaimoNiskanen:
There are also some hyphenations such as "use-case", "record-like", "single-quote" that I think should be written as separate words.

The Markdown linter complains about mixing fenced code blocks (` ```erlang `) with indented code blocks, that our mandatory header looks like / happens to be.  Ignore those, and I will try to find how to disable that warning.

Other than that, and my inline comments, I think it looks good!

RaimoNiskanen:
> @RaimoNiskanen, I did not include you as an author on the EEP, but since you implemented the new record syntax, you obviously deserve to be on it. I just didn't want to include you without your blessing first.

Although I finished the reference implementation, and thus more or less defined what this EEP should describe, I'd say you are still the author of the EEP, also despite me having a bunch of review comments...

RaimoNiskanen:
It seems I accidentally pressed a bad button and closed this... Oops!  Sorry about that!

RaimoNiskanen:
I have updated the workflow configuration (on 'master') and re-run it, but the unwanted warnings are still there.  Can you try to rebase this PR to the latest 'master'?

eeps/eep-0072.md
```diff
@@ -0,0 +1,145 @@
...
+definitions (also adopting the above syntactical changes), such that the
+following record definitions would all be valid and identical:
+
+```erlang
+-record('div', {a :: integer(), b :: integer()}).
+-record #'div'{a :: integer(), b :: integer()}.
+-record #div{a :: integer(), b :: integer()}.
+```
+
+Motivation
```
robertoaloi, line 28: The motivation section covers the reason behind the usage of unquoted "keywords" atoms, but not the eventual benefits from the introduction of the new record syntax. I can see how the latter could be more opinionated than the former. Maybe we should limit this EEP to the usage of unquoted atoms only and, if there's really a benefit from multiple syntaxes for record definitions, have a separate EEP/motivation?

RaimoNiskanen: The motivation I'd say is to be able to use the same quoting for record definition as for usage.  Otherwise one would have to know which atoms that has to be quoted in the record definition, which is awkward and error prone, or always quote the atom in the record definition, which is ugly.

But I agree that this should be in the Motivation chapter.

There should also be a motivation for why we want to allow variable names unquoted, and that is for applications such as ASN.1 or Corba that have such name conventions and that now have to quote the record names just like for the `'div'` example.

Separating into two different EEP:s I think would be unfortunate, because we then might get a period of asymmetry when one cannot use the same format in record definition and usage.

That there are so many variants of how to define a record is largely due to the syntax rules for attributes and atom quoting in Erlang.

Maybe we should instead write that the old and the new syntaxes are:

```erlang
-record('div', {a :: integer(), b :: integer()}).
-record #div{a :: integer(), b :: integer()}.
```

and then elaborate that due to parentheses being optional in attributes, and due to that atoms may be quoted even when not needed, these happen to be equivalent, just mentioned for completeness:

```erlang
-record 'div', {a :: integer(), b :: integer()}.
-record(#div{a :: integer(), b :: integer()}).
-record #'div'{a :: integer(), b :: integer()}.
-record(#'div'{a :: integer(), b :: integer()}).
```

michalmuskala:
While I do understand the arguments for using unquoted atoms in record syntax and I consider it a simple and relatively small change, the change to record definition syntax is a major one and I believe should go through a bigger consideration.

In particular, the argument for symmetry, even with the change to definition, still doesn't hold since things like `is_record/2` guard won't work with unquoted names. So, you'd still need to know which names require quoting and which don't.

Furthermore, the EEP completely misses to discuss the impact on language by having 2 separate and parallel syntaxes for defining records. This will be confusing to learners and will inherently cause some split in "best practices" and across codebases. This might be OK if benefits outweigh this cost, but as it stands, I am not convinced this is true.

RaimoNiskanen:
> In particular, the argument for symmetry, even with the change to definition syntax, still doesn't hold since things like is_record/2 guard won't work with unquoted names. So, you'd still need to know which names require quoting and which don't.

That is correct.  The symmetry argument is only for definition vs. usage, which I think is the important ones.  Everybody has to define a record to be able to use it.

Record names are atoms, and by `is_record/2` and the other record meta functions this shines through...  At some point in the language we will have to draw the line where it shines through, if we as this EEP proposes, relax the name rules for record usage.

I have always been a bit annoyed that record definition looks different from record usage, and from when defining a record type (where the name also can be unquoted, in this EEP:s reference implementation).  Here I saw an opportunity to fix that, for convenience.

The old record definition style would be obsolete, and I don't think there is any reason to keep using it other than for backwards compatibility reasons.

choptastic:
> _edit_ I just got enlightened that hyphen-composed words are used in adjective-like ways, by many writers of American English. So "record-like" above is by that OK.

:heart:  Fascinating! I love it. I absolutely adore learning things that I do every day or take for granted only to discover they are either local, regional, or national patterns. I had no idea that heavy hyphen use in this way was an American thing. I will review my overuse of hyphenated non-adjective-like words and revise accordingly :)

And thanks everyone for all the comments and discussions.  I'm working on further revisions to the EEP and hope to have them pushed within the next hour or so!

choptastic:
Thank you everyone for your comments and concerns.

I've expanded the EEP to include a few things:

1. Actually [expanded the motivations and specs](https://github.com/erlang/eep/pull/64/files#diff-bd86e0152371d3afe4379c20461b5c85467ac474a2d3f9c7603d1528a7b5a560R131-R174) for the new definition syntax 
2. Acknowledged @michalmuskala's concerns regarding the [impact](https://github.com/erlang/eep/pull/64/files#diff-bd86e0152371d3afe4379c20461b5c85467ac474a2d3f9c7603d1528a7b5a560R206-R213) the new definition syntax may have on the language going forward and I at least [acknowledged](https://github.com/erlang/eep/pull/64/files#diff-bd86e0152371d3afe4379c20461b5c85467ac474a2d3f9c7603d1528a7b5a560R201-R205) the still-present inconsistency between the new syntax and existing functions (good call-out).
3. Expanded the motivations to include the [variable-caps atoms](https://github.com/erlang/eep/pull/64/files#diff-bd86e0152371d3afe4379c20461b5c85467ac474a2d3f9c7603d1528a7b5a560R100-R106)

Perhaps my approach with having two motivations and two specifications is a bit nontraditional, however, I do address the one leads to the other (as @RaimoNiskanen) discovered.

Thanks again everyone - I was genuinely surprised to see so much discussion on this when I woke up this morning (
I originally submitted this just after midnight then went to sleep). You folks are *on top of it*! :)

erszcz:
> The old record definition style would be obsolete, and I don't think there is any reason to keep using it other than for backwards compatibility reasons.

Maybe that's too much to ask... but maybe not? 🙏 If the EEP implementation came with an accompanying utility for converting existing code from the old syntax to the new one, this compatibility reason could become redundant soon. I mean a tool similar to what Lukas Larsson provided for the EDoc -> Markdown transition.

eeps/eep-0072.md
RaimoNiskanen: You have a point. Maybe just "like variable names".  I interpret "atom with variable capitalization" as `'Hello'` because quoting is how to create such an atom.
We should avoid inventing an acronym...

RaimoNiskanen:
> 4. I rebased from the new master so that hopefully the workflow stuff works now?

There are still some Markdown linter warnings.  Mostly that there should be blank lines surrounding fenced code blocks and headings, and that there should not be unneeded consecutive blank lines.

choptastic:
I'll try to clean it up and see if that fixes the issues.

Thanks!

RaimoNiskanen:
> I'll try to clean it up and see if that fixes the issues.

One last remaining on line 46.

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+The latter two are proposed new syntax.  The following would also be valid
+and identical since parentheses are optional in attributes:
+
+```erlang
+-record 'div', {a :: integer(), b :: integer()}.
+-record(#'div'{a :: integer(), b :: integer()}).
+-record(#div{a :: integer(), b :: integer()}).
+```
```
RaimoNiskanen, line 35: I think the intention to add *one* new format would be better shown by just having `-record('div', {` and `-record #div{` in the first group, and write that due to optional quoting and optional parentheses the second group (of 4) are identical and equivalent.

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+foo_Bar.
+'foo-bar'. % still quoted since the term has a non-atomic character in it.
+```
+
+Conveniently, this also means that records named with simple atoms can be
+invoked and used without having to quote the atoms. For example:
+
+```erlang
+-record(foo, {a, b}).
+
```
RaimoNiskanen, line 68: I am trying to squeeze in a capital letter record further down, but is not sure this works out well enough...  It is a suggestion.

```suggestion
-record(bar, {c}).

```

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+go() ->
+    X = #foo{a = 1, b = 2},
+    Y = X#foo{a = something_else},
+    Z = Y#foo.a,
```
RaimoNiskanen, line 72:
```suggestion
    Z = #bar{c = Y#foo.a},
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+    ...
+```
+
+Unfortunately, that also means that records named with anything that doesn't
+fit the "simple atom" pattern must be wrapped in quotes in definition and
+usage. For example:
+
+```erlang
+-record('div', {a, b}).
+
```
RaimoNiskanen, line 82:
```suggestion
-record('SET', {c}).

```

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+go() ->
+    X = #'div'{a = 1, b = 2},
+    Y = X#'div'{a = something_else},
+    Z = Y#'div'.a,
```
RaimoNiskanen, line 86:
```suggestion
    Z = #'SET'{c = Y#'div'.a},
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+1. Allowing reserved words and variables to be used without quotes for record
+   names, and
+2. Simplifying the definition such that the syntax between record definition
+   and record usage becomes more consistent.
+
+With the changes from this EEP, the above code becomes:
+
+```erlang
+-record('div', {a, b}).
+
```
RaimoNiskanen, line 124:
```suggestion
-record('SET', {c}).

```

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+go() ->
+    X = #div{a = 1, b = 2},
+    Y = X#div{a = something_else},
+    Z = Y#div.a,
```
RaimoNiskanen, line 128:
```suggestion
    Z = #SET{c = Y#div.a},
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+    ...
+```
+
+Implementation
+==============
+
+To update the syntax for using records, we can safely augment the parser to
+change `# atom` to `# record_name` and define `record_name` to be either
+`atom`, `var`, or `reserved_word`.
```
RaimoNiskanen, line 182:
```suggestion
change its already existing record handling of `'#' atom '{' ... '}'` and `'#' atom '.' atom` into `'#' record_name '{' ... '}'` and `'#' record_name '.' atom`, and define
`record_name` to be `atom`, `var`, or `reserver_word`.
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+
+To update the record definition syntax, we can simply add a few new
+modifications to the `attribute` Nonterminal.
```
RaimoNiskanen, line 185:
```suggestion
modifications to the `attribute` Nonterminal to allow `'#' record_name` as name for the `record` attribute, instead of `atom` as for generic attributes.
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+
+1. Other functions that work with records like `is_record/2` or `record_info/1`
+   are not currently covered by any of the syntactical changes in this EEP, and
+   as such, it remains necessary to quote record names if they are not simple
+   atoms. For example: `is_record(X, div)` would still be a syntax error. So
+   there is still not true 100% symmetry.
+2. This EEP introducing a new syntax for record definition could potentially
+   lead to some to wonder why the language has two rather different syntaxes
+   for defining records. Since usage of the syntax for getting, setting,
+   matching, etc (e.g. `#rec{a=x,y=b}`) occurs far more commonly than defining,
```
RaimoNiskanen, line 211: I would like to add type definition to this list, since it is another annoying discrepancy that for a type definition of a record type the record syntax is used, but as of today not for the record definition itself.

eeps/eep-0072.md
```diff
@@ -0,0 +1,227 @@
...
+analyzing code using the new syntax.
+
+Syntax highlighting and code completion tools may need to be updated to support
+the new syntax if your code uses the new syntax rules.
+
+Broader Concerns and Points of Discussion
+=========================================
+
+While the new definition syntax creates some degree of symmetry around record
+usage, there do remain some concerns and inconsistencies:
```
RaimoNiskanen, line 201:
```suggestion
usage, perfect symmetry is impossible to achieve, since a record can always
be handled as the atom tagged tuple it actually is. The question is where
to draw the line where the record's true nature shows, and how hard we
should try to hide it. These are remaining concerns and inconsistencies:

File: eeps/eep-0072.md [hunk 19]
```diff
@@ -0,0 +1,227 @@
...
+
+1. Other functions that work with records like `is_record/2` or `record_info/1`
+   are not currently covered by any of the syntactical changes in this EEP, and
+   as such, it remains necessary to quote record names if they are not simple
+   atoms. For example: `is_record(X, div)` would still be a syntax error. So
+   there is still not true 100% symmetry.
```
RaimoNiskanen, line 207:
```suggestion
   there is still not true 100% symmetry.  Note that instead of using the
   `is_record(X, 'div')` guard, matching on `#div{}` is probably more frequently used,
   since it is terser and mostly regarded as more readable.

File: eeps/eep-0072.md (hunk 8 above)
Comment (choptastic) [reply eeps/eep-0072.md:35]: Yes, that's right - I forgot to adjust this in my round of updates.  I will adjust accordingly.

File: eeps/eep-0072.md (hunk 17 above)
Comment (choptastic) [reply eeps/eep-0072.md:220]: That's a great point! I'll fit it in somewhere here.

File: eeps/eep-0072.md [hunk 20]
```diff
@@ -0,0 +1,250 @@
...
+========
+
+This EEP loosens some of the restrictions around record names to make it
+no longer necessary to quote them when they are named with reserved
+words (`#if` vs `#'if'`) or words with capitalized first characters (`#Hello`
+vs `#'Hello'`).
+
+This EEP also proposes to add a new record-like syntax to the record
+definitions (also adopting the above syntactical changes), so that the
+following record definitions would all be valid and identical:
```
RaimoNiskanen, line 20: Nitpick:

```suggestion
following record definitions would be valid and identical:
```

eeps/eep-0072.md
```diff
@@ -0,0 +1,250 @@
...
+
+```erlang
+-record('div', {a :: integer(), b :: integer()}).
+-record #div{a :: integer(), b :: integer()}.
+```
+
+The latter one is proposed new syntax.  The following would also be valid
+and identical since parentheses are optional in attributes:
```
RaimoNiskanen, line 28: Nitpick:

```suggestion
and identical since parentheses are optional in attributes,
and since atoms may be quoted even when not mandatory:
```

choptastic (commented):
Merge updated phrasing

eeps/eep-0072.md
choptastic: I've updated the phrasing here: https://github.com/erlang/eep/pull/64/commits/dbfb39e626bc930f6378af45c1375c2599955edb

What do you think?

choptastic:
Alrighty, sorry for not commenting about my pushes last week.  The days got away from me.

I've pushed a few small updates, but I think it's all good now, but let me know if you think I've missed anything big.

RaimoNiskanen (approved):
I think this looks fine!

RaimoNiskanen:
The OTP Technical Board meeting decided to postpone this proposal since we have other record improvement possibilities on the radar, and don't want to accidentally shut a syntactical door on such an improvement.  We'll see for OTP 29 how these syntax extensions can play in, they will definitely be considered or incorporated.

choptastic:
@RaimoNiskanen, A related improvement might be to allowing the attribute names to follow the new proposed syntax as well, thereby allowing attribute names to be `if`, `Variable`, etc.

I haven't done anything to attempt to do this, but might that also be one of the things the team is considering?  Given my work on Nitrogen is so heavily bound to records, I'm all kinds of intrigued by more improvements to working with records.

PR #73, opened 2025-01-07
Priority Messages

rickard-green:
In some scenarios it is important to propagate certain information to a process quickly without the receiving process having to search the whole message queue which can become very inefficient if the message queue is long. This EEP introduces the concept of priority messages to the language which aim to solve this issue.


eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+
+The reference implementation can be found in [pull request 9269][] of the
+[Erlang/OTP repository][].
+
+Care has been taken to have as small impact as possible on processes not
+utilizing priority messages. Processes not enabling reception of priority
+messages will not use any more memory at all due to the priority messages
+implementation.
+
+A few Notes on the Implementation
```
bjorng, line 296:
```suggestion
A Few Notes on the Implementation
```

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+has used priority messages in a way not intended.
+
+### Priority Messages in Transit
+
+There exists a number of different types of signals. For each type of signal an
+action is taken when the signal is received. Ordinary messages are special
+since they are very common and the only action taken upon reception of an
+ordinary message is to add it to the end of the message queue. Due to this,
+the signal queue for incoming signals is arranged as a skip list where each
+non ordinary message signal points to the next non ordinary message signal.
```
bjorng, line 349:
```suggestion
non ordinary message signal points to the next non-ordinary message signal.
```

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+This way we can move a whole batch of ordinary messages into the message queue
+at once.
+
+Priority marked message signals need to be sent as non ordinary message
```
bjorng, line 353:
```suggestion
Priority marked message signals need to be sent as non-ordinary message
```

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+signals, since they need to have another action taken than the default. There
+are other signals that are received as non ordinary message signals, but then
```
bjorng, line 355:
```suggestion
are other signals that are received as non-ordinary message signals, but then
```

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+transformed into ordinary messages depending on the state of the receiving
+process. An example of such a signal is a message sent using an alias. Upon
+reception of such a message the receiver checks if the alias is still active.
+If it is, then adds it to the end of the message queue; otherwise, drops the
```
bjorng, line 359:
```suggestion
If it is, then adds it to the end of the message queue; otherwise, it drops the
```

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+handle *alternate action messages*. Both a priority marked message and a
+message sent using an alias are just messages with an alternate action to take
+upon reception than the default, so both of them will use the alternate action
+message implementation.
+
+[Message Reception]: eep-0076-1.png
+    "Message Reception"
+
+[Priority Message Reception]: eep-0076-2.png
+    "Priority Message Reception"
```
bjorng, line 371: Could you redraw the figures to get rid of the wiggly lines under words such as "Msg" and "PrioMsg"?

bjorng (approved):

josevalim (approved):

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+message queue lengths of processes in the system. When a message queue length
+exceeds a certain limit, you might want to change strategy of handling incoming
+messages. In order to do that, you typically need to inform the process with a
+long message queue about this.
+
+Sending it a message informing about the long message queue will not work,
+since this message will end up at the end of the long message queue. If the
+receiver handles messages one at a time in message queue order, it will take a
+long time until the receiver fetch this message. The situation will at this
+point very likely have become even worse.
```
Maria-12648430, line 54: Or the situation could have resolved itself by then (like, if the messages occur in sudden bursts), at which point it would be too late to change handling strategies.

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+  messages due to a specific monitor being triggered by calling the
+  `process_flag/2` BIF like this:
+  `process_flag({priority_monitor_message, MonitorRef}, true)`.
+  The receiver can also opt-in for reception of priority monitor messages by
+  passing the option `priority` in the option list that is passed as third
+  argument to the `monitor/3` BIF when creating the monitor.
+
+The receiver process can at any time disable reception of certain priority
+messages by passing `false` as second argument to any of the above listed
+`process_flag/2` BIF calls.
```
Maria-12648430, line 188: If a sending process specified in a priority process flag exits, will it automatically be removed from the process flags of the receiver?

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+extended in the future if that should be the case. The following list
+describes how the different types of messages will be enabled as priority
+messages:
+
+* *Priority Marked Messages* - A message is marked as a priority message by the
+  sender by passing the option `priority` in the option list that is passed as
+  third argument to the `erlang:send/3` BIF. The receiver opts-in for reception
+  of priority marked messages from a specific sender by calling the
+  `process_flag/2` BIF like this:
+  `process_flag({priority_marked_message, SenderPid}, true)`.
```
Maria-12648430, line 173: I wonder, how could this be put to work for `system` messages originating from `sys` and by extension from `proc_lib`? One would probably want to prioritize those, but AFAICT the sending process is the calling or a freshly spawned process, ie it can't be known to set the `SenderPid` for the `priority_marked_message` flag.

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+
+By letting priority messages overtake ordinary messages that already exist in
+the message queue we get priority messages ordered in reception order among
+priority messages followed by ordinary messages ordered in reception order
+among ordinary messages. Instead of just overtaking ordinary messages, one
+could choose to let a priority message overtake all messages in the message
+queue regardless of whether they are priority messages or not, but then
+multiple priority messages would accumulate in reverse order. Having these two
+sets of messages ordered internally by reception order at least to me feels the
+most useful. Just as in the case of ordinary messages we will probably want to
```
Maria-12648430, line 140: IMO, keeping reception order is important, imperative, the only sensible thing to do. More so since the receiving process can't distinguish prioritized from ordinary messages.

Maria-12648430 (commented):
Pretty cool, I was hoping for something like this for a while :heart:

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+handle priority messages in reception order.
+
+Note that the reception order of signals is not changed. If a process sends an
+ordinary message and then a priority message to a another process, the ordinary
+message will be received first and then the priority message will be received.
```
mikpe, line 145: "received" here doesn't mean "returned by `receive`" I take it, since the whole point of this proposal is to change that order?

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+  argument to the `monitor/3` BIF when creating the monitor.
+
+The receiver process can at any time disable reception of certain priority
+messages by passing `false` as second argument to any of the above listed
+`process_flag/2` BIF calls.
+
+The reason for not having options for accepting all priority marked messages,
+all exit messages, or all monitor messages as priority messages is the risk of
+introducing bugs when code in other modules are called from the process
+accepting priority messages. For example, if a process enables all monitor
```
mikpe, line 193: I think this point should be emphasized. Disabling the message receive order guarantees should be opt-in and based on explicit agreement between sender and receiver (which may be the same module).

eeps/eep-0076.md
Maria-12648430: Certainly, but it leads to an interesting point.

If the order of events turns out like this...
1. process A sends ordinary message M to process B
2. process A sends prio message P to process B
3. process B calls `receive`
4. process B calls `receive`
... then process B will `receive` the messages in order `P` then `M`.

However, if the order of events turns out like this...
1. process A sends ordinary message M to process B
2. process B calls `receive`
3. process A sends prio message P to process B
4. process B calls `receive`
... then process B will `receive` the messages in order `M` then `P`.

To be clear, that is pretty much as expected. It could/should be mentioned maybe.

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+queue than where the receive marker was inserted. One solution to this problem
+could be to disable the selective receive optimization on processes that
+enables priority messages. The user of priority messages would in that case
+have to be very careful not to call into modules that might rely on the
+selective receive optimization. This would more or less make it impossible to
+safely call modules that you don't have full control over yourself, since it
+in the future might be modified in a way so that it relies on the selective
+receive optimization taking effect. Therefor I find it unacceptable to disable
+the selective receive optimization. The priority message implementation needs
+to be able to preserve the selective receive optimization.
```
mikpe, line 236: I'd say MUST not needs.

eeps/eep-0076.md
Maria-12648430: I guess when you say "the same module" you really mean "the same _process_"? :sweat_smile:

eeps/eep-0076.md
```diff
@@ -0,0 +1,394 @@
...
+messages, and selected monitor messages should be treated as priority messages.
+Perhaps one would want other types of messages to be treated as priority
+messages as well, but the set of allowed priority messages can easily be
+extended in the future if that should be the case. The following list
+describes how the different types of messages will be enabled as priority
+messages:
+
+* *Priority Marked Messages* - A message is marked as a priority message by the
+  sender by passing the option `priority` in the option list that is passed as
+  third argument to the `erlang:send/3` BIF. The receiver opts-in for reception
```
Maria-12648430, line 170: Should there be an `!`-like operator for sending messages prioritized (like, say, `!!`)? Probably not, having a convenient operator may lead to abuse and/or over-use. Also, it would be somewhat hard to factor hypothetical priority levels into it.

eeps/eep-0076.md
mikpe: No, I mean the same module. Say a `gen_server` where an API function (called from outside the server process) ends up passing a priority message to the server process -- the code for both sender and receiver is in the same module.

Maria-12648430: Hm, ok, after re-reading the rest of the paragraph I get that. However, the significance of the module from which a call is made is unclear to me. Why does it matter? :thinking:

mikpe: The only significance is that with both sender and receiver in the same module they are in a better position to agree on the validity of using priority messages, which otherwise would be totally invalid.

My point is that priority messages break long-standing Erlang semantics and are _totally_ invalid in the general case. They must be opt-in, which requires a close relationship between sender and receiver.

eeps/eep-0076.md
```diff
@@ -0,0 +1,395 @@
...
+  of priority marked messages from a specific sender by calling the
+  `process_flag/2` BIF like this:
+  `process_flag({priority_marked_message, SenderPid}, true)`.
```
potatosalad, line 174: Would it be possible to allow something like this for allowing any process on the current node to send priority messages to the receiving process?

```erlang
process_flag({priority_marked_message, local}, true).
```

Otherwise, for cases like overload protection where it may be a system-wide decision to change strategies, it could require having a secondary process used solely for the purpose of forwarding high priority messages to the receiving process based them opting in to receive priority messages from this secondary proxy process.

essen:
If you send priority messages fast enough, the process may never get to read normal messages again. This warrants a big warning in the documentation, or to make the implementation read from both queues. Considering the intended use case, the former sounds good enough.

eeps/eep-0076.md
essen: I don't think we want to encourage the use of priority messages *that much*. They should be rare enough to not need an operator.

eeps/eep-0076.md
rickard-green: In the reference implementation this is currently not done, but this can be implemented.

eeps/eep-0076.md
rickard-green: > I wonder, how could this be put to work for `system` messages originating from `sys` and, by extension, from `proc_lib`?

To me it feels a bit like stretching what the priority messages was intended for

rickard-green: > So, as a first idea, maybe `priority_marked_message` should accept aliases too?

Hmm, I like it. Perhaps the `priority` option can be replaced with *priority aliases* all together.

eeps/eep-0076.md
rickard-green: > "received" here doesn't mean "returned by `receive`" I take it, since the whole point of this proposal is to change that order?

Correct. The signal is received before it is inserted into the message queue. The name of the `receive` expression is in my opinion quite misleading. A less misleading name (although not as nice name) could have been `fetch_message_from_message_queue`.

eeps/eep-0076.md
rickard-green: I've rephrased this.

rickard-green:
> If you send priority messages fast enough, the process may never get to read normal messages again. This warrants a big warning in the documentation, or to make the implementation read from both queues. Considering the intended use case, the former sounds good enough.

Yes, but I'd argue that you are misusing the feature if that should happen and quite a lot of features can destroy a system if misused (for example process priorities). Agree about the warning.

eeps/eep-0076.md
potatosalad: Double checking my understanding, that idea might look like this, right?

```erlang
% receiver
PriorityAlias = erlang:alias([priority]).
% distribute to "senders" out of band

% sender (any process)
PriorityAlias ! priority_message_here.
```

rickard-green: @potatosalad Yes

eeps/eep-0076.md
```diff
@@ -0,0 +1,395 @@
...
+
+Alternative Solutions Considered
+--------------------------------
+
+A separate priority message queue per process exposed to the Erlang program
+could be an alternative solution. You would need a way similar to this
+proposal to choose which messages should be accepted as priority messages.
+There would also need to be some new syntax in order to multiplex matching of
+messages from the different message queues. This would be a larger change of
+the language without providing any extra benefits as I see it.
```
michalmuskala, line 254: As I suggested in https://github.com/erlang/otp/pull/8371#issuecomment-2081554865, a separate mailbox could lead a way to a very trivial implementation of a worker pool - multiple workers receiving from a shared mailbox. In general, MPMC queues would be quite useful, as they are widely used in the industry (assuming current mailboxes are MPSC queues).
Arguably, it also makes the ordering semantics much simpler - the order is maintained within a mailbox, and switching mailboxes for a process would be explicit.

juhlig: Interesting idea... but also a _massive_ change, not only in terms of behind-the-scenes changes but also in how a user would have to think about messages (or signals) in general. Generally, taking into account "making message queues a first-class object" as said in [erlang/otp#8371 (comment)](https://github.com/erlang/otp/pull/8371#issuecomment-2081554865), it would amount to "process B can "steal" messages from the inbox of process A", something that you definitely have to get used to after years (decades?) of being an Erlang user.

If you want to have MPMC queues, baked into the language or not, I think they _have_ to be something distinctly different from process message (signal) queues. Ie, you can't replace one with the other, at least not without massive repercussions.

I guess what I'm trying so say is that the fact that it does make a _particular_ use case trivial as you put it, doesn't mean that it is a _generally_ good idea, in the wider sense. Put differently, it is IMO a different idea suited for a different use case, and therefore outside of the scope of this EEP.

eeps/eep-0076.md
josevalim: I dropped a longer comment on the Erlang Forum but I worry that relying on runtime behaviour, such as aliasing and distribution of said aliases, to control the priority messages may complicate correctness analysis in the future, or even general understanding of the code. I'd generally strive to make it clearer both in the sender and the receiver that a given message is a priority.

eeps/eep-0076.md
essen: If you're going to have a shared mailbox that you need to switch to, then it can just be a new module that takes care of all implementation details. Today we have processes and tables; tomorrow we could have a third type of object that is a mailbox, or a queue, that multiple processes could use. It could have semantics similar to Erlang process mailboxes, only shared (you wouldn't replace RabbitMQ with it, but it could help in various node-local scenarios). I don't think it requires special syntax, only considerations around how best to get messages from these queues.

eeps/eep-0076.md
essen: I don't think it's too far fetched to expect system messages to be handled optionally as priority messages. Imagine you are debugging an overloaded process and you are trying to get the process state. Today you must be very patient. With priority system messages you can immediately debug things. If I understand correctly aliases would enable that scenario (since system messages are ultimately normal messages that could be prioritised).

rickard-green:
I've updated the EEP. The `process_flag/2` flags for identifying messages to handle as priority messages has been replaced by *priority aliases* and `priority` options.

PR #74, opened 2025-01-10
Draft: comprehension assignment

richcarl:
This EEP adds assignments to comprehension qualifier lists.
Reference implementation PR: https://github.com/erlang/otp/pull/9153

### DIFF
--- a/eeps/eep-0077.md
+++ b/eeps/eep-0077.md
@@ -1,0 +1,121 @@
+    Author: Richard Carlsson <carlsson.richard(at)gmail(dot)com>
+    Status: Draft
+    Type: Standards Track
+    Created: 10-Jan-2025
+    Erlang-Version: OTP-28.0
+    Post-History:
+****
+EEP 77: Assignment in Comprehensions
+----
+
+Abstract
+========
+
+This EEP adds assignments to comprehension qualifier lists, providing
+a convenient and readable alternative to other syntactical tricks.
+
+Rationale
+=========
+
+It would often be useful to be able to easily bind variables in the
+qualifier sequence of a comprehension, for example:
+
+```erlang
+[Z || X <- Ls,
+      {foo, Y} = g(X),
+      Z <- f(Y, h(Y))].
+```
+
+using plain `Pattern = ...,` entries between qualifiers.
+
+You can do this today by writing a singleton generator:
+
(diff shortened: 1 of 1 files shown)


eeps/eep-0076.md
```diff
@@ -0,0 +1,118 @@
...
+It is in fact already allowed syntactically to have a `Pattern = ...`
+match expression in the qualifiers. This however gets interpreted as
+any other expression - thus expected to produce a boolean value - and
+if `false`, the current element will be skipped.
+
+Hence, any qualifier following after a match `Var = Expr` will only be
+executed if `Var` has the value `true`. We can therefore expect that
+no such uses exist in practice, because `Var` would be fully
+redundant. (The OTP code base has been checked and does not contain
+any.)
```
josevalim, line 86: I'd say this is still a backwards incompatible change, because if someone relies on it, the behaviour will change in very subtle ways:

```
2> [{X, Y} || X <- [1, 2.0, 3], Y = is_integer(X)].
[{1,true},{3,true}]
```

Also, for completeness, one could introduce maybe expressions' `?=`operator in comprehensions, which would work similar to an assignment + filter (as it will discard the element if it does not match the pattern).

richcarl: It is an incompatible change, yes. To protect against that, one could start with making `Pat = ...` an error in OTP 28, and then add the new interpretation in OTP 29. I really did not want to add yet another special token. (`?=` could be useful though). One possibility is to use the so far reserved token 'let' as a prefix: `let Pat = ..., ...`.

michalmuskala: There's an option of making it an error in 28, and introducing an opt-in feature that uses the new semantics (rather than erroring), and in 29 making this feature the default.
This way, it won't silently give errors, but won't delay the new syntax for another release for people that are ready for it.

josevalim: Sounds like a winner to me.

kikofernandez:
We will review this in the near future.

Thanks for your contribution

kikofernandez:
@richcarl We will wait a bit to see how the community (Erlang Forums and interested people reading this EEP) sees this change. Thanks

kikofernandez (approved):

PR #75, opened 2025-02-03
EEP 78: Multi-valued comprehensions

michalmuskala:
**TL;DR**: This EEP proposes enhancing the comprehension syntax to allow emitting multiple
elements in a single iteration of the comprehension loop - effectively enhancing
comprehensions to implement `flatmap` with a fixed number of elements, for example:

```
[X + 1, X + 2, ... || X <- Xs]
```

### DIFF
--- a/eeps/eep-0078.md
+++ b/eeps/eep-0078.md
@@ -1,0 +1,114 @@
+    Author: Michał Muskała <micmus(at)whatsapp(dot)com>
+    Status: Draft
+    Type: ...
+    Created: 16-12-2024
+    Erlang-Version: ...
+    Post-History: ...
+****
+EEP 78: Multi-valued comprehensions
+----
+
+Abstract
+========
+
+This EEP proposes enhancing the comprehension syntax to allow emitting multiple
+elements in a single iteration of the comprehension loop - effectively enhancing
+comprehensions to implement `flatmap` with a fixed number of elements.
+
+Rationale
+=========
+
+Comprehensions in Erlang are a very flexible and convinent way of implementing many
+iteration and looping patterns. However, there are some cases that end up unergonomic,
+notably, and where the scope of this EEP is, where we'd like to emit multiple elements
+from a single iteration.
+
+Existing forms of expressing this are awkward and introduce extra, unnecessary allocations.
+For example:
+
+    lists:append([[X + 1, X + 2] || X <- Xs]
+    [Tmp || X <- Xs, Tmp <- [X + 1, X + 2]]
+
+Both of those ways end up creating an extra allocation of a temporary 2-element list, introducing
+inefficiency, as well as are, arguably, harder to understand than necessary introducing extra function
+calls or variables.
+
+This EEP proposes enhancing comprehensions with the ability to do so using a very natural
+syntax extension of the existing comprehension syntax. In particular for list and map
+comprehensions:
+
+    [X + 1, X + 2, ... || X <- Xs]
+
+    #{K + 1 => V + 1, K + 2 => V + 2, ... || K := V <- Map}
+
+The semantics of map comprehensions where multiple keys in the same iteration would end up
+with the same value, should be the same as if the keys were emitted in subsequent iterations.
+
+Binary comprehensions already support this, and thus there's no enhancement to their syntax
+suggested in this EEP, for example:
+
+    << <<(X + 1), (X + 2)>> || <<X>> <= Bin>>.
+
+Parsing & Abstract Forms
+==============
+
+Today, the [comprehension abstract forms](https://www.erlang.org/doc/apps/erts/absform.html#expressions) are defined as:
+
+- If `E` is a list comprehension `[E_0 || Q_1, ..., Q_k]`, where each `Q_i` is a
+  qualifier, then `Rep(E) = {lc,ANNO,Rep(E_0),[Rep(Q_1), ..., Rep(Q_k)]}`. For
+  `Rep(Q)`, see below.
+- If `E` is a map comprehension `#{E_0 || Q_1, ..., Q_k}`, where `E_0` is an
+  association `K => V` and each `Q_i` is a qualifier, then `Rep(E) =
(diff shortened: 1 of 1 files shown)


michalmuskala:
The first expression will be an error:
```
test.erl:6:31: variable 'Z' is unbound
%    6|   [ Z = binary_to_integer(X), Z || X <- [<<"1">>, <<"2">>, <<"3">>] ].
%     |                               ^
```
Similar how it would be in a plain list `[X = binary_to_integer(~"1"), X]`.

The second will indeed yield as you expected:
```
3> [ hello, goodbye || true ].
[hello,goodbye]
```

The third will again error with `Hello` variable undefined:
```
test.erl:6:29: variable 'Hello' is unbound
%    6|   #{hello => Hello = hello, Hello => goodbye || true }.
%     |                             ^
```
The same would happen if you tried this with a plain map `#{hello => Hello = hello, Hello => goodbye}` won't compile.

In other words, similar to all "containers" expressions are evaluated "in parallel" not "in sequence" as far as variable scoping is concerned.

Maria-12648430:
Hm, there is an unconvenient limitation, namely that it is not possible to decide if something should be inserted once, twice, thrice etc, or not at all (yes, that can and still has to be done with filters). That is, with the proposed syntax, it is only possible to add elements a fixed number of times, which IMO puts a severe limitation on the ergonomics the extension would otherwise provide.

That said, I also find it somewhat confusing. It looks like a set of expressions where, everywhere else, the _last_ would be the result, but here the result of _each_ expression is inserted.

Maria-12648430:
> I guess you should think of it as the list of expressions at the "head" of a cons operator…

Ok, makes sense :)

eeps/eep-0078.md
```diff
@@ -0,0 +1,114 @@
...
+
+    #{K + 1 => V + 1, K + 2 => V + 2, ... || K := V <- Map}
+
+The semantics of map comprehensions where multiple keys in the same iteration would end up
+with the same value, should be the same as if the keys were emitted in subsequent iterations.
+
+Binary comprehensions already support this, and thus there's no enhancement to their syntax
+suggested in this EEP, for example:
+
+    << <<(X + 1), (X + 2)>> || <<X>> <= Bin>>.
```
Maria-12648430, line 50: This is true, but also the multi-value syntax should be applicable to it, if only for the sake of consistency and completeness.

I think a binary comprehension like...
```erlang
<< <<(X + 1)>>, <<(X + 2)>> || <<X>> <= Bin >>.
```
... should be accepted and result in a concatenation of the generated binaries, that is, same as if written `<<(X + 1), (X + 2)>>`.

michalmuskala: Yes, that's a good point. TBH the syntax for binary comprehensions is quite noisy already and I didn't want to muddle it more

kikofernandez:
Hi @michalmuskala , thanks for your contribution. 
Did you create a topic in the ErlangForums? I did not see, but maybe I was not looking too much in depth

michalmuskala:
@kikofernandez I've created one now: https://erlangforums.com/t/eep-78-multi-valued-comprehensions/4537

TD5:
The proposal as stated by @michalmuskala would be incredibly useful for me, since I have a lot of hot-code which ends up with precisely the style/performance trap described with:

```
lists:append([[X + 1, X + 2] || X <- Xs]
```

or

```
[Tmp || X <- Xs, Tmp <- [X + 1, X + 2]]
```

> Hm, there is an unconvenient limitation, namely that it is not possible to decide if something should be inserted once, twice, thrice etc, or not at all (yes, that can and still has to be done with filters). That is, with the proposed syntax, it is only possible to add elements a fixed number of times, which IMO puts a severe limitation on the ergonomics the extension would otherwise provide.

I am not sure about an arbitrary number of elements, but for a closed-set of elements to be prefixed, I wonder whether something akin to this is possible:

```
[
  case X of
    one -> X + 1;
    two -> X + 1, X + 2;
    three -> X + 1, X + 2, X + 3
  end
|| X <- List]
```

Of course, this precise syntax is not a good idea at all, since there's now a strange new kind of expression in the tail position of a case-expression, but presumably we could devise some syntax for inside comprehension to conditionally determine the elements to emit.

Maria-12648430:
> Of course, this precise syntax is not a good idea at all, since there's now a strange new kind of expression in the tail position of a case-expression, but presumably we could devise some syntax for inside comprehension to conditionally determine the elements to emit.

It's not a new kind of expression, this is valid today, with the last expression of a case branch being the one that gets inserted in the list. Which is definitely worse in more than one regard than strange new expressions.

That said, I am also not a huge fan of introducing brand new syntaxes that only can be used in this particular context of comprehensions. I would argue that with things getting as complex as this, it is out of the scope of what is at the end of the day a shorthand syntax, and explicit recursion or a fold should be used.

kikofernandez (approved):

richcarl:
As @TD5 noted, this is limited to static repetitions, and the more general and common pattern that it would be nice to address is when you generate varying sublists to be joined, as in `lists:append([f(X) || X <- ... ])`. If there existed an appending comprehension (using e.g. `<|>`), it could be written `[f(X) <|> X <- ... ]`. The question is however whether we can gain any efficiency. Unless we can hope for deforestation/stream fusion optimization to kick in (unlikely), there will still need to be intermediate sublists under the hood as return values from f(X). At some point these sublists must be traversed to build the final result. Making a temporary list of lists and appending them at the end is in fact among the cheapest possible ways to implement this - like IOlists and scatter/gather in general. (The fastest is to build the sublists in reverse and join them up with `lists:reverse/2` - also known as "reverse onto", if you really must have all the speed.)

Suppose you write a "deforested" version: if it's recursive it will need to allocate a stack frame per element to hold intermediate data, and if it's tail recursive it will need to build an accumulator list and reverse it in the end. The amount of work is much the same as the appending version - the only real difference would be a shorter syntax (with yet another symbol `<|>`). So to summarize: the EEP's suggestion is simple and can be efficient, though limited, but maybe it's the best we can do. For the general case, don't fear the append.

PR #79, opened 2025-04-22
Update EEP-47 to reflect non-centralized proposal

garazdawi:
This commit changes the error index to be non-centralized and also irons out a lot of the details on how it is expected to be used with various tools.

/cc @robertoaloi

### DIFF
--- a/eeps/eep-0074.md
+++ b/eeps/eep-0074.md
@@ -1,36 +1,40 @@
     Author: Roberto Aloi <prof3ta(at)gmail(dot)com>
+            Lukas Backström <lukas(at)erlang(dot)org>
     Status: Accepted
     Type: Standards Track
     Created: 11-Nov-2024
     Erlang-Version: OTP-28
-    Post-History:
+    Post-History: Updated April 2025 to use non-centralized index
 ****
-EEP 74: Erlang Error Index
+EEP 74: Erlang Diagnostic Index
 ----
 
 Abstract
 ========
 
-The **Erlang Error Index** is a _catalogue_ of errors emitted by
-various tools within the Erlang ecosystem, including - but not limited
-to - the `erlc` Erlang compiler and the `dialyzer` type checker.
+The **Erlang Diagnostic Index** is a standardized way to _catalogue_
+errors/warnings/information messages emitted by various tools and applications
+within the Erlang ecosystem, including - but not limited to - the `erlc`
+Erlang compiler, the `dialyzer` type checker and the `ssl` application.
 
-The catalogue is not limited to tools shipped with Erlang/OTP, but it
-can include third-party applications such as the [EqWAlizer][]
-type-checker or the [Elvis][] code style reviewer.
+The indexes are not limited to Erlang/OTP, but can also be used by third-parties
+such as the [Elixir] language, the [EqWAlizer] type-checker or the [Elvis]
+code style reviewer.
 
-Each error in the catalogue is identified by a **unique code**
-and it is accompanied by a description, examples and possible courses
-of action. Error codes are _namespaced_ based on the tool that
-generates them. Unique codes can be associated to a human-readable
-**alias**.
+Each diagnostic in an index is identified by a **code** and it is accompanied
+by a description, examples and possible courses of action. Diagnostic codes are
+_namespaced_ based on the tool that generates them.
 
-Unique error codes can be leveraged by IDEs and language servers to
-provide better contextual information about errors and make errors
-easier to search and reference. A standardized error index creates a
-common space for the Community to provide extra examples and
-documentation, creating the perfect companion for the Erlang User
-Manual and standard documentation.
+The namespaces for diagnostic indexes are not maintained centrally, but
+left up to the community to co-ordinate. It is recommended that one
+searches online before taking a namespace, just as when deciding on an
+application name.
+
+Diagnostic codes can be leveraged by IDEs and language servers to
+provide better contextual information about errors and warnings and
+make them easier to search and reference. A standardized diagnostic index
+creates a common way for the community to provide extra examples and
+documentation.
 
 Rationale
 =========
@@ -43,31 +47,31 @@ Producing meaningful error messages can sometimes be challenging for
 developer tools such as compilers and type checkers due to various
 constraints, including limited context and character count.
 
-By associating a **unique code** to each _diagnostic_ (warning or
-error) we relief tools from having to condense a lot of textual
-information into a - sometime cryptic - generic, single
+By associating a **unique code** to each _diagnostic_ (warning,
+error or information) we relieve tools from having to condense a
+lot of textual information into a - sometime cryptic - generic, single
 sentence. Furthermore, as specific wording of errors and warnings is
-improved over time, error codes remain constant, providing a
+improved over time, diagnostic codes remain constant, providing a
 search-engine friendly way to index and reference diagnostics.
 
-An good example of this is the _expression updates a literal_ error
-message, introduced in OTP 27. Given the following code:
+A good example of this is the _expression updates a literal_ warning
+message, introduced in Erlang/OTP 27. Given the following code:
 
     -define(DEFAULT, #{timeout => 5000}).
 
     updated(Value) ->
       ?DEFAULT#{timeout => Value}.
 
-The compiler emits the following error:
+The compiler emits the following warning:
 
     test.erl:8:11: Warning: expression updates a literal
         %    8|   ?DEFAULT#{timeout => 1000}.
         %     |           ^
 
-The meaning of the error may not be obvious to everyone. Most
+The meaning of the warning may not be obvious to everyone. Most
 importantly, the compiler provide no information on why the warning is
 raised and what a user could do about it. The user will then have to
-recur to a search engine, a forum or equivalent to proceed.
+refer to a search engine, a forum or equivalent to proceed.
 
 Conversely, we can associate a unique identifier to the code (say,
 `ERL-1234`):
@@ -76,7 +80,7 @@ Conversely, we can associate a unique identifier to the code (say,
         %    8|   ?DEFAULT#{timeout => 1000}.
         %     |           ^
 
-The code make it possible to link the error message to an external
+The code makes it possible to link the warning message to an external
 resource (e.g. a wiki page), which contains all the required,
 additional, information about the error that would not be practical to
 present directly to the user. Here is an example of what the entry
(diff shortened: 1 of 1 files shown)


eeps/eep-0074.md
```diff
@@ -76,34 +80,293 @@ Conversely, we can associate a unique identifier to the code (say,
...
+more examples and references to the documentation.
+
+````
+
+For example:
+
+````markdown
+# ERL-0001 - Function head mismatch
+
+## Error example
```
robertoaloi, line 285: Maybe avoid "Error" here.

eeps/eep-0074.md
```diff
@@ -76,34 +80,293 @@ Conversely, we can associate a unique identifier to the code (say,
...
+
+```erlang
```
robertoaloi, line 287: Any way we could leverage doctests here?

eeps/eep-0074.md
```diff
@@ -43,31 +47,31 @@ Producing meaningful error messages can sometimes be challenging for
 developer tools such as compilers and type checkers due to various
 constraints, including limited context and character count.
 
-By associating a **unique code** to each _diagnostic_ (warning or
-error) we relief tools from having to condense a lot of textual
-information into a - sometime cryptic - generic, single
+By associating a **unique code** to each _diagnostic_ (warning,
+error or information) we relieve tools from having to condense a
```
robertoaloi, line 51: Add the "hint" severity or simply drop the listing.

robertoaloi (approved):
LGTM, only left a couple of minor nits/questions.

eeps/eep-0074.md
```diff
@@ -1,36 +1,40 @@
...
+The indexes are not limited to Erlang/OTP, but can also be used by third-parties
+such as the [Elixir] language, the [EqWAlizer] type-checker or the [Elvis]
+code style reviewer.
 
-Each error in the catalogue is identified by a **unique code**
-and it is accompanied by a description, examples and possible courses
-of action. Error codes are _namespaced_ based on the tool that
-generates them. Unique codes can be associated to a human-readable
-**alias**.
+Each diagnostic in an index is identified by a **code** and it is accompanied
```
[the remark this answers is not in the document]
garazdawi:
```suggestion
Each diagnostic in an index is identified by a uniqye **code** and it is accompanied
```

I meant to remove the "human-readable alias" part, not the unique part. Is the above better?

eeps/eep-0074.md
```diff
@@ -1,36 +1,40 @@
...
 ----
 
 Abstract
 ========
 
-The **Erlang Error Index** is a _catalogue_ of errors emitted by
-various tools within the Erlang ecosystem, including - but not limited
-to - the `erlc` Erlang compiler and the `dialyzer` type checker.
+The **Erlang Diagnostic Index** is a standardized way to _catalogue_
+errors/warnings/information messages emitted by various tools and applications
```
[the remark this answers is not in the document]
garazdawi: How many levels are enough levels will always be hard to know. We added 7 levels to logger and I think we only use 3 in Erlang/OTP :)

I'll try to update the text to relax the wording.

eeps/eep-0074.md
```diff
@@ -1,36 +1,40 @@
...
-Unique error codes can be leveraged by IDEs and language servers to
-provide better contextual information about errors and make errors
-easier to search and reference. A standardized error index creates a
-common space for the Community to provide extra examples and
-documentation, creating the perfect companion for the Erlang User
-Manual and standard documentation.
+The namespaces for diagnostic indexes are not maintained centrally, but
+left up to the community to co-ordinate. It is recommended that one
+searches online before taking a namespace, just as when deciding on an
+application name.
```
okeuday, line 31: You appear to claim Ericsson should determine all namespaces used with Erlang errors, is that correct?  Why not add a namespace concept to the Erlang programming language, to avoid a central naming bottleneck?

okeuday: Later on in the document namespaces are defined as examples for Erlang/OTP source code that creates an association between the namespace name and the application name like `ERL` == `compiler` as shown in the output of `application:get_diagnostic("ERL-0001")`.  The situation would be a lot simpler if we could just use application names as namespaces.  Why add a separate name as a namespace that still must be associated with an application name?

okeuday: If the namespace name lacks visibility, it only makes the diagnostic code more obscure and difficult to utilize (e.g., to discover its context with search engine results).  I understand the goal of having a shorter prefix for diagnostic codes but see it as creating new problems, instead of improving diagnostic codes.  The new problems are 1) how to avoid conflicts with all Erlang source code everywhere, 2) how to associate the diagnostic code prefix with an Erlang/OTP application name that relates to the release process, any repository storage, and its existence on the filesystem (the directory path could contain the application name, the application file name and its contents, and the application's module name prefixes).  Erlang/OTP application names are typically short and easy to remember, so they provide a good reference to find any mention of that source code and documentation.

Using the word "namespace" is also a problem because I was immediately wanting an application name module prefix.  I understand that idea has been avoided for decades and the old package idea in Erlang was removed because Erlang/OTP systems are generally kept small to avoid any concerns regarding namespaces and multiple versions of the same dependencies.

Since it is difficult to anticipate use it should be best to make diagnostic codes easy to reference and that should require the use of application names for diagnostic code prefixes, because all Erlang source code is typically referenced based on an application name.

essen: [My proposal to use URIs](https://github.com/erlang/eep/pull/70#issuecomment-2469990326) fell on deaf ears in the previous PR. It would have simplified a lot of things.

garazdawi: > [My proposal to use URIs](https://github.com/erlang/eep/pull/70#issuecomment-2469990326) fell on deaf ears in the previous PR. It would have simplified a lot of things.

I did not use the URI proposal for the same reason as I don't like using the application name as a prefix, they become too long. You can shorten it by using some sort of translation scheme as you proposed, but are you not then doing the same thing as this proposal?

essen: > I did not use the URI proposal for the same reason as I don't like using the application name as a prefix, they become too long. You can shorten it by using some sort of translation scheme as you proposed, but are you not then doing the same thing as this proposal?

The key difference is that your method requires you to search for the diagnostics, whereas the URI scheme defines where the diagnostics are found.

Your URI scheme can have well defined components, and some of those components can carry additional information (for example what https://en.wikipedia.org/wiki/Well-known_URI does). If you had a URI such as `erl:elvis/abc` the first segment of the path could refer to a class of error. Additional information about the class of errors could be found in a registry, and for each class of error the registry entry would have information on where to find diagnostics (locally or online, using URI templates for easy translation).

The registry information can be embedded directly in tools that require it. It isn't expect to change often. It could come directly with Erlang/OTP, even if some of its information is about third party applications.

Depending on context the `erl:` scheme can be optional (just like `http:` in HTTP URIs), it's just that it can be useful to automatically detect them and produce links to online diagnostic information. But if you really want to, `elvis/0001` or `ELV/0001` is perfectly fine to print, and not longer than what you already got.

There's certainly more details to figure out but that's the rough idea.

garazdawi: > The key difference is that your method requires you to search for the diagnostics, whereas the URI scheme defines where the diagnostics are found.

It does not require you to search. You can get from ERL-0001 to https://erlang.org/doc/ERL-0001.html without searching, but I realize that is not clear enough in the document. The suggestion is that the new application key `documentation_url` can be combined with the diagnostic code to create that url. You do need access to the application that generated the diagnostic in order to do the translation though. The `application:get_diagnostic/1` example shows it:

```erlang
> application:get_diagnostic("ERL-0001").
{ok, [#{application => compiler,
        filename => "/home/erlang/lib/compiler/doc/diagnostics/ERL-0001-update-literal.md",
        url => "https://erlang.org/doc/ERL-0001-update-literal.html",
        short => "ERL-0001",
        long => "ERL-0001-update-literal",
        doc => ~"# ERL-0001\n..."}]}
```

essen: Yes you need to have the application in the path and you need to search for the correct application, this is what I mean by "searching". You do not need either with a URI scheme, you just need access to the registry, and you can embed that registry much easier than having to include all the tools.

Think of a use case where you'd check for errors coming from CI logs for example. In your case you need to have all applications that produce error codes in your path. With my proposal you can just have a small escript do the job.

You could add a registry to your proposal as well as parsing rules for the error codes and whatnot, but you'd just be reinventing the wheel and potentially making the diagnostic class and error format a lot stricter than they should be. If I was using this in Cowboy it would make a lot more sense if I could have something like `cow/<proto>/<code>` for example, possibly with the `<code>` part being a word or two considering the number of possible errors. You could still provide recommendations for keeping them short without requiring it.

garazdawi: > you just need access to the registry

I think the word "just" here is doing a lot of work. In my proposal you just need access to the application that generated the diagnostic.

In my opinion we should strive to avoid a centralized registry of any kind, especially one that we embed into Erlang/OTP. If we do end up wanting a central registry, then hex is the closest thing we have in the Erlang world and then we are back at @okeuday's reasoning that we should use application names in the diagnostic codes to make translation trivial. We could work with the hex team to allow redirects to happen from <https://hex.pm/packages/cowboy/diagnostics/*> to <https://ninenines.eu/docs/en/cowboy/2.13/manual/*>, or just accept that applications without documention on hexdocs.pm will not get URIs.

As an aside, the idea of the `documentation_url` comes from a request of ExDoc users to be able to use autolinks to documentation not on hexdocs.pm. I'm just piggybacking on that idea.

> Think of a use case where you'd check for errors coming from CI logs for example. 

If you are parsing the output then you should probably want the `json` output from the EEP and in there you have the URL already. But there are bound to be some scenarious when you cannot use `json` and then the you may have to resort to using your favourite search engine or LLM.

essen: A registry is just a table of data. The only centralized part about it is that there has to be a single source of truth for it. Beyond that, it's just data.

A script could have the data from the registry embedded directly in the code, or have the parts of the registry it cares about handled as code directly. It doesn't have to ship with OTP in its entirety, although it could if useful. In practice I suppose OTP would only include the data it needs from the registry.

The EEPs are a kind of centralized registry that OTP team manages. The diagnostics class registry would be similar, only much simpler as each entry only needs a small number of fields. Here's an example of registry: https://www.iana.org/assignments/http-fields/http-fields.xhtml - a table. For diagnostics the fields would likely be the diagnostic class e.g `erlc` which must be unique, the application or local URI for local lookup, the web URI for online viewing and perhaps some one sentence description. You can manage the registry as a single text file with changes done via PRs, you don't really need any more than that. Could be an EEP that gets updated as necessary. Hex is very far from this idea.

Regardless of the solution you go for you will need to ensure that the diagnostic class is unique, and a registry helps with that.

garazdawi: I just realized that one reason for not using a registry is because it would make it impossible to do a diagnostic code to URL mapping for private/proprietary projects.

It would also be possible to scrape the projects on hex.pm and create a hex application for that registry to make it available to everyone (except Erlang/OTP that is).

> Regardless of the solution you go for you will need to ensure that the diagnostic class is unique, and a registry helps with that.

Why would we need that? We don't have any such mechanism for many things in Erlang (applications, modules, process names etc etc) and it has worked ok for now? From time to time there have been collisions, but any collisions that would take place can be solved by the people building the applications involved. I imagine the same would happen here.

essen: A registry doesn't prevent private projects to do diagnostic code to URL mapping, a registry is just data and private projects can provide their own data for the mapping. Just like there are application-specific "proprietary" HTTP headers that aren't in the HTTP header registry (GRPC has some for example). In our case, tools that read diagnostics just need a way to be fed the private data.

Proprietary projects can register their own diagnostic class and mappings even if fewer people would encounter them. I would imagine diagnostics to be useful to QuickCheck for example, and can't think of a reason to prevent them from registering.

> It would also be possible to scrape the projects on hex.pm

It is much simpler to maintain a text file with the occasional PR. There aren't that many projects that would truly benefit from diagnostics. Even its use in Cowboy would be a stretch, I'm only considering it because I have things like this all over the place: https://github.com/ninenines/cowlib/blob/master/src/cow_http2_machine.erl#L381-L383 which could be replaced by diagnostics like `cow/h2/data_when_idle`. But there aren't too many non-CLI projects that have this amount of error details around.

> Why would we need that?

That one was a general "you", not OTP team. Someone creating a new diagnostic class needs to make sure there's no conflicts, and a registry is a good start to do that.

Thanks for digging into the idea in any case!

garazdawi:
We are merging the current version for now. There will probably be updates to this document before the final implementation is in place.

PR #81, opened 2025-11-06
Add EEP for native records

bjorng:


josevalim (commented):
Thank you @bjorng and @garazdawi! The proposal is well written and building on top of records does a beautiful job of keeping the language changes minimal. Removing the `undefined` as default for fields is another great change, as well as having the optional of having "required fields" (which must be given on creation).

My only criticism to the proposal is that it doesn't discuss maps at all. For the last ~10 years, we have been using maps as replacements for records, and while native records aim to ease the migration between "records -> native records", there are no paths between "maps -> native records". Perhaps this is intentional, you are not expecting anyone to migrate between maps and native records, but then we have to deal with the side-effect that, as native records improve, a lot of code (specially in Elixir) will be forever suboptimal.

If I am allowed to spitball a bit, I'd love if "native records" were implemented as "named maps" or "record maps" behind the scenes. The proposal would stay roughly the same, the only differences are that:

* Accessing a field in any record could use the existing map syntax and just work:

  ```erlang
  Expr#Field
  ```
  
* Updating any record could use the existing map syntax and just work:

  ```erlang
  Expr#{Field1=Expr1, ..., FieldN=ExprN}
  ```

You could also support the proposed `#_{...}` syntax and it would additionally check it is a record and not a "plain map". `is_map/1` would obviously return true for records but you could do a more precise check with `is_record`.

Regarding the key-destructive map operations, such as `maps:put/3` or `maps:remove/2`, I'd make them fail unless you "untag" the map (or you could allow them to succeed but the tag would be removed at the end, which I find too subtle).

Overall, the goal would be to effectively unify records and maps, removing the decade-old question "records or maps". This would also provide a path for Erlang and Elixir to unify behind the same constructs, so I'd love to hear your opinions.

eeps/eep-0079.md
```diff
@@ -0,0 +1,763 @@
...
+
+```erlang
+#user{name = ~"John", city = ~"Stockholm"}
+```
+
+The next expression is used to create a new user native record outside
+of the defining module:
+
+```erlang
+#misc:user{name = ~"John"~, city = ~"Stockholm"}
```
essen: Extra ~?

eeps/eep-0079.md
```diff
@@ -0,0 +1,763 @@
...
+characters. Fields can be in any order.  The compiler validates that
+all `Field1`, ..., `FieldN` are unique, and if defined in the same
+module it will issue a warning if keys are not present.
+
+#### Default values
+
+If no value is provided for a field, and there is a default field
+value in the native record definition, the default value is used. If
+no value is provided for a field and there is no default field value
+then a native record creation fails with a `badrecord` error.
```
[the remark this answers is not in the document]
essen: More context definitely better, even if it gives only one missing key out of several.

eeps/eep-0079.md
```diff
@@ -0,0 +1,763 @@
...
+
+* The field `Field` is not defined in the native-record value.
+
+The expression can be used in guards — a guard would fail if the
+corresponding expression raises.
+
+The following syntax allows accessing field `Field` in any record:
+
+```erlang
+Expr#_.Field
```
essen: This creates a new use case because we could have two records `v1` and `v2` that share some fields, and use this syntax to access the common fields. It probably needs to be highlighted.

essen (commented):
Great work!

Would it make sense to have finer grained control on who can do what? For example restrict creation to the defining module while still providing access to fields; or read-only fields outside the defining module. Probably doesn't matter for 95% of use cases I reckon.

eeps/eep-0079.md
bjorng: Yes, will fix.

eeps/eep-0079.md
```diff
@@ -0,0 +1,769 @@
...
+All of these checks are straightforward, and `xref` can be easily extended to handle them.
+
+### Performance characteristics
+
+While the implementation of native-records is a large scope, from the
+runtime perspective they are much closer to maps than to records
+(tuples). So, their performance characteristics should align with maps
+(with insignificant overhead for runtime validation). Additionally,
+given that native-records are more specialized versions of maps (with
+all keys being atoms), there is potential for optimizations.
```
lpil: Is the expectation that native records will have inferior performance to tuple records?

lpil (commented):
What a cool EEP! Thank you! I've a handful of questions

# Language interop

One aspect of this which the document does not touch on that I think could be highly impactful for the BEAM ecosystem is language interop. Today each major language has a different preference for fixed key-value data structures:

- Erlang: maps and records
- Elixir: maps with a special field containing a module atom
- Gleam: records

This creates some degree of friction when calling code from other BEAM languages. If they all were to largely use native records then this friction go away, making interop between languages would be a much better experience.

I'm not immediately seeing any problems for Gleam as we use records there, but it seems like it would be more challenging for Elixir there we use maps.

# Adoption within existing OTP modules

It seems that in an ideal world that native records would be the ubiquitous data structure, once they are available. Would existing Erlang/OTP modules be updated to work with them?

Functions that expect classic records, tagged tuples, and maps could have new function clauses added to handle native records in a backwards compatible way, unless I am mistaken. It seems that due to not being compatible with maps or tuples there would be very little ability to update existing functions to return native records.

Is there something we could do here? Or is the expectation that Erlang/OTP code will use different data structures depending on how old it is?

# Construction syntax

Is the only difference between the native and classic record syntaxes the `#` character in the name of the definition? This seems like it will be very error prone, and also hard for less familiar people to debug as the definition will be accepted by the Erlang compiler, but their attempts to construct the record will fail.

---

Thank you all!

bjorng:
> Functions that expect classic records, tagged tuples, and maps could have new function clauses added to handle native records in a backwards compatible way, unless I am mistaken. It seems that due to not being compatible with maps or tuples there would be very little ability to update existing functions to return native records.

Yes. If a function returns a tuple record, all we can do is to create a new function that returns a native record.

> Or is the expectation that Erlang/OTP code will use different data structures depending on how old it is?

To some extent, yes. I think that is already the case.

> If they all were to largely use native records then this friction go away, making interop between languages would be a much better experience.

Agreed.

lpil:
Thank you @bjorng.

To confirm: all fields must have names? There are no positional fields?

Can one define a record which does not have any fields?

```erlang
-record #none{}.
-record #some{value = term()}.
-type option() :: #none{} | #some{}
```

eeps/eep-0079.md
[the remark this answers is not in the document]
bjorng: I've changed the error reason to `{novalue,Field}` (where `novalue` is a new error reason).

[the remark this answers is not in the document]
dgud: > Accessing a field in any record could use the **existing** map syntax and just work:
>    Expr#Field

In erlang there is **no** existing map syntax for accessing a single field today,
that is still missing, was never implemented I guess because maps can have arbitrary terms as keys.

I guess it could be added for "literal" fields? 
And it is really annoying to have to use the add record_name when accessing fields in records today, e.g. Rec#record_name.field

dgud:
> Great work!
> 
> Would it make sense to have finer grained control on who can do what? For example restrict creation to the defining module while still providing access to fields; or read-only fields outside the defining module. Probably doesn't matter for 95% of use cases I reckon.

We are trying to limit the scope here to try to get it in to 29. 

Hmm, this would require some additional syntax, 'private' | 'protected' | 'public' (borrowing from ets access types), personally
I don't think this is necessary nor wanted, Opaqueness on fields have been up for discussion but dropped for now at, implementation and reflection reasons.

eeps/eep-0079.md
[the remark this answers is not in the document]
bjorng: > I am assuming that, as long as you pattern match on the named record when the variable is defined, the compiler would be able to optimize "unamed" field access and updates?

Yes, that should be possible.

essen:
> Hmm, this would require some additional syntax, 'private' | 'protected' | 'public' (borrowing from ets access types), personally I don't think this is necessary nor wanted, Opaqueness on fields have been up for discussion but dropped for now at, implementation and reflection reasons.

Right it's something that likely doesn't have to be done at runtime, but it is important to consider at least in the documentation part, as internal fields definitely shouldn't be documented. Read-only fields can be marked as such easily in the text. But this depends on what the documentation will look like I suppose.

bjorng:
Update: there can now be two distinct errors when `Rec#rec.field` fails: `{badrecord,Term`} or `{badfield,field}`.

RaimoNiskanen:
@tsloughter wrote:
>Related to not replacing records, I think -record being overloaded is confusing. ...

"Not replacing records" is really phrased: not "Replacing all tuple record usage scenarios.", and "Replace most tuple-record usages without having to update anything but the declaration.".

With that in mind I think overloading -record may be more good than bad.

potatosalad:
### Question: Removing a field

Consider this scenario:

**Node A (old code):**

```erlang
-record #config{host, port, legacy_timeout}.
Config = #config{host = "localhost", port = 8080, legacy_timeout = 5000}
```

**Node B (new code - legacy_timeout removed):**

```erlang
-record #config{host, port}.
```

When Node A sends `Config` to Node B:

1. The removed field still exists in the value: The record value contains `[host, port, legacy_timeout]` because that's what Node A created.
2. Reading removed fields: Based on the spec, `Config#config.legacy_timeout` should still work on Node B because field access doesn't consult the definition—it only checks if the field exists in the value. This means:
  - Code on Node B can accidentally read fields that "don't exist" in its definition
  - Linters/dialyzer on Node B would flag `legacy_timeout` as an error, but it works at runtime
  - Might this create a confusing situation for developers?
3. Pattern matching works: `case Config of #config{host = H, port = P} -> ... end` works fine (only matching fields that exist in both value and definition).

**Issue:** The interaction between field access (which ignores definition) and pattern matching (which may or may not check definition) is unclear.

eeps/eep-0079.md
```diff
@@ -0,0 +1,784 @@
...
+### Native-record guard BIFs
+
+#### `is_record/2`
+
+We define a pseudo-BIF `is_record/2`, which will be translated by
+the compiler to instructions to test whether `Term` is a native
+record.
+
+```erlang
+is_record(Term :: dynamic(), #Module:Name) -> boolean().
```
richcarl: I really don't like the idea of allowing free `#r` or `#m:r` as in `is_record(X, #r)`, because those fragments do not have any semantic meaning on their own, but they will have to be handled as something that may occur in any expression position. I think the rationale further below is wrong in this case: it's much worse to introduce pseudo-expressions that have no defined meaning outside a specific context. And you won't be using that syntax for the API functions `records:create(Module, Name, ....)` etc. anyway.

eeps/eep-0079.md
```diff
@@ -0,0 +1,784 @@
...
+
+Yes, we should add a BIF that essentially does the following:
+
+```erlang
+record:get_fields(Record) =:= record:get_fields(#Module:Record{})
+```
+
+but guaranteed to always work and be more efficient.
+
+TODO: What should the name of the BIF be?
```
richcarl: Something like is_subtype or is_compatible, maybe.

josevalim: I think subtype and compatible would make me think the type of the fields are the same and are also checked. Perhaps `has_record_fields` but there are no guards starting with `has_`. I am not sure how to phrase it using `is_`.

eeps/eep-0079.md
```diff
@@ -0,0 +1,784 @@
...
+  an attempt is made to match out a a value from a non-exported record
+  from a module other than its defining module.
+
+* The pattern references a FieldK and the native-record value does not
+  contain this field.
+
+Note, however, that it is possible to match on the name of a
+non-exported record. Thus, if the `match_name/1` function in the
+following example is called with an instance of record `r` defined in
+`some_module`, it will succeed even if the record is not exported:
```
richcarl: What is the motivation for this matching on non exported records by name?

eeps/eep-0079.md
```diff
@@ -0,0 +1,802 @@
...
+The following items will never be considered as goals.
+
+1. Replacing **all** tuple record usage scenarios.
+2. Allowing variable key lookup.
+3. Allowing variables as a record name.
+4. Supporting creation of records in guards.
+5. Supporting `element/2` for native records.
+6. Supporting opaque records.
+7. Having `undefined` as default when neither the record definition nor
+   the record creation provides a value a field.
```
rickard-green:
```suggestion
   the record creation provides a value for the field.
```

eeps/eep-0079.md
```diff
@@ -0,0 +1,802 @@
...
+Field names must be atoms.
+
+When a native-record value is updated, its native-record
+definition is **not** consulted.
+
+An update operation fails with a `{badrecord,Expr}` error if:
+
+* When the native-record value was created, the definition of the
+  record was not exported, and it is now used outside its definining
+  module.
```
rickard-green:
```suggestion
* The definition of the record was not exported when the native-record
  value was created, and it is now used outside its defining module.
```

eeps/eep-0079.md
```diff
@@ -0,0 +1,802 @@
...
+< record()
+< map()
+< []
+< [_|_]
+< bitstring()
+```
+
+Native-record values are ordered by their fully qualified name, then
+by their visibility, then their keys, and finally by field values
+(field values are considered as maps). Equality is defined through
```
rickard-green: What does "field values are considered as maps" mean?

bjorng: I will remove that part.

eeps/eep-0079.md
```diff
@@ -0,0 +1,798 @@
...
+There is no defined maximum number of fields in a native record. Because of
+the need to define all fields in a record, it is very hard to accidentally
+create a record with one million elements.
+
+### `-import_record()`
+
+As it can be seen from the next sections, working with native records outside of
+the defining module needs using fully qualified names of the native records:
+`#misc:user{}`, ... This quickly may become too cumbersome and verbose. The
+`-import_record` directive works similar to the `-import` and `-import_type` directives.
```
tazjin: Why do we need explicit directives for each case, instead of folding e.g. import_record into one of the existing ones? Is this an explicit design choice, or a technical restriction?

tazjin:
Good proposal! My overall thought is that I see this as a step towards "cleaning up" struct-like types, i.e. types with known fields. Strongly agree with @josevalim on the unification of maps and records due to this.

In fact, this non-goal from the proposal irks me a bit:

> Replacing **all** tuple record usage scenarios.

This can be read in two ways:

1. Native records shouldn't support all concrete uses in code that tuple records have (which makes sense, e.g. for plain pattern matching on the underlying tuples, which would obviously be unsupported).
2. Native records do not aim to be "better" at all use-cases that tuple records are good at.

I think that being better should be a design goal: Native records look, in many ways, like "Records V2" - which sets an expectation (and, imho, a reasonable one) that they should be strictly *better* than "Records V1".

I also think introducing as little new syntax and definitions as possible for this would also be great. If there was a way to just declare a record in essentially "the map type language", maybe literally through the `-type` keyword as somebody suggested above, *and* get additional compiler optimisations from that - it would be amazing.

yarisx:
When reading about public | private access of fields in another comment I've got a question regarding native records hierarchy:
- as I understand it is possible to define a native record that has a field which itself is a native record (although there should be some work when creating values of such type)
- if so, it is possible (intentionally or accidentally) to export the "outer" record definition without exporting the "inner" record definition
What will be the behaviour of code manipulating instances of the "outer" record type? As I can understand, as long as the code does not try to manipulate the field(s) of the "inner" record everything should be fine, including pattern-matching on the name of the "inner" record (but not its fields).

UlfNorell:
This looks very nice! I would like to suggest a
```erlang
-spec to_map(record()) -> #{atom() => term()}.
```
in the reflection API to make it easy to write generic traversals supporting native records:
```erlang
foo(BaseCase) -> ...;
foo([H|T]) ->
  [foo(H)|foo(T)];
foo(M) when is_map(M) ->
  maps:from_list(foo(maps:to_list(M));
foo(R) when is_record(R) ->
  records:create(records:get_module(R), records:get_name(R), foo(records:to_map(R));
foo(X) -> X.
```

eeps/eep-0079.md
```diff
@@ -0,0 +1,798 @@
...
+An access operation fails with a `{badfield,Field}` error if:
+
+* The field `Field` is not defined in the native-record value.
+
+The expression can be used in guards — a guard would fail if the
+corresponding expression raises.
+
+### Anonymous access of native records
+
+The following syntax allows accessing field `Field` in any record:
```
Maria-12648430: > The following syntax allows accessing field `Field` in **any record**:

Any record... does this mean it also applies to non-native records?

eeps/eep-0079.md
```diff
@@ -0,0 +1,798 @@
...
+6. Supporting opaque records.
+7. Having `undefined` as default when neither the record definition nor
+   the record creation provides a value for the field.
+8. `_` wildcard default for assigning a value to any field not explictly
+   initialized.
+
+Soft Non-Goals
+--------------
+
+Here are some features we might consider implementin in a future release.
```
zachallaun:
```suggestion
Here are some features we might consider implementing in a future release.
```

eeps/eep-0079.md
```diff
@@ -0,0 +1,798 @@
...
+#### Default values
+
+If no value is provided for a field, and there is a default field
+value in the native record definition, the default value is used. If
+no value is provided for a field and there is no default field value,
+a native record creation fails with a `{novalue,FieldName}` error.
+
+In OTP 29, the default value can only be an expression that can be
+evaluated to a constant at compile-time. Essentially, the allowed
+expressions are guard expression, with the following additional
```
zachallaun:
```suggestion
expressions are guard expressions, with the following additional
```

eeps/eep-0079.md
```diff
@@ -0,0 +1,798 @@
...
+    * Native records can travel between the Erlang nodes having different
+      versions of code.
+
+From now on let’s employ more precise terminology (when needed):
+*native-record definitions* and *native-record values*.
+
+When a native-record value is created, it “captures” key information
+from the current native-record definition, namely:
+
+* They fully qualified name of the native-record (module name and
```
zachallaun:
```suggestion
* The fully qualified name of the native-record (module name and
```

zachallaun (commented):
Excited to see this! Comments are only related to minor typos found during my read-through.

eeps/eep-0079.md
```diff
@@ -0,0 +1,798 @@
...
+```erlang
+Expr#Name.Field
+Expr#Module:Name.Field
+```
+
+These expressions return the value of the specified field of the
+native-record value.
+
+When a field from a native-record value is accessed, the current
+native-record definition is **not** consulted.
```
mikpe: Can I assume this means that accesses only use information contained in or referenced by the native record value itself, and there is no lookup to the current version of the native record's creating module?

I understand that implementation details are outside the scope of a language proposal, but it would be useful to have (or link to) a discussion about possible implementations. Understanding the performance implications of native records compared to maps and tuple records would be essential before changing code to use them.

filmor:
Do you already have an idea on how to expose these to NIFs?

bjorng:
> Do you already have an idea on how to expose these to NIFs?

I expect similar to add terms, but we haven't looked in the details yet.

brigadier:
Will it be possible to get list of keys from either the record definition or a bound record variable?

bjorng:
I've pushed a number of `fixup` commits, attempting to clarify unclear things.

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+-import_record(example1, [user, pair]).
+```
+
+### Definitions
+
+NOT_REVIEWED_BY_OTB.
+
+From now on let’s employ more precise terminology (when needed).
+
+A *native-record defintion* (or *record definition* for short)
```
josevalim:
```suggestion
A *native-record definition* (or *record definition* for short)
```

eeps/eep-0079.md
```diff
@@ -0,0 +1,769 @@
...
+  module).
+
+* The field `Field` is not defined in the native-record value.
+
+The expression can be used in guards — a guard would fail if the
+corresponding expression raises.
+
+### Anonymous access of native records
+
+The following syntax allows accessing field `Field` in any record:
```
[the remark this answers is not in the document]
josevalim: @bjorng I assume the optimizations could still be applied for the anonymous syntax if you matched on the full syntax elsewhere before (basically when beam_types will have it narrowed down to a given the record type)?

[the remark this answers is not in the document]
bjorng: Yes, the compiler can still do optimizations if previous matching has narrowed down the type. (When we'll actually implement type-analysis for native records, which we haven't done yet.)

The optimizations I was thinking of was the optimizations in the runtime system planned for OTP 30, which I've outlined under "Performance characteristics". Anonymous records will be handled similarly to small maps, which will still be quite efficient.

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+
+A *native-record definition* (or *record definition* for short)
+is the definition of a record in some module:
+
+```erlang
+-module(example).
+-record #person{name, age}.
+```
+
+A *native-record value* (or *record value* for is the term created by
```
RaimoNiskanen: Typo: ... `(or the *record value* for short)` ...

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+with the definition in the module. However, there are some
+circumstances in which this is not the case. For example:
+
+* If the code for the module has been unloaded.
+
+* If a version of the module with a different native record definition
+  has been loaded.
+
+* If a record value was sent to another node where the code for the
+  module was not loaded.
```
RaimoNiskanen: Maybe add something like: `, or the record definition is different in the loaded module on that node.`

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+In OTP 29, the default value can only be an expression that can be
+evaluated to a constant at compile-time. Essentially, the allowed
+expressions are guard expressions, with the following additional
+restrictions:
+
+* No variables.
+
+* No calls of any kind.
+
+* No creation of any kind of records.
```
RaimoNiskanen: Not even tuple records?  Why not?

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+native record definition. It fails with a `badrecord` error in the
+following cases:
+
+* There is no corresponding native-record definition.
+
+* The native-record definition is not visible at the call site (it is
+  not exported).
+
+* The native record create expression references the field FN which is
+  not defined (in the structure definition).
```
RaimoNiskanen: Suggestion:
```
... references a field FN which is
not defined (in the native record definition).
```

I think there are more references to "structure definition", that probably should be changed into "native record definition".

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+```erlang
+Expr#Name.Field
+Expr#Module:Name.Field
+```
+
+These expressions return the value of the specified field of the
+native-record value.
+
+When a field from a native-record value is accessed, the
+native-record definition captured is **not** consulted.
```
RaimoNiskanen: Is that even correct?  I thought:
```
When a field from a native record value is accessed,
its **captured** definition is consulted.
```

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+
+An access operation fails with a `{badrecord,Expr}` error if:
+
+* The definition of the record was not exported when the native-record
+  value was created, and it is now used outside its definining
+  module.
```
RaimoNiskanen: Suggested rewrite for (in my mind) more logical order of explanation:
```
* The native-record value's captured definition states it as
  from a different module than the calling module, and *not exported*.
```
There are more like this...

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+
+```erlang
+Expr#Name{Field1=Expr1, ..., FieldN=ExprN}
+Expr#Module:Name{Field1=Expr1, ..., FieldN=ExprN}
+```
+
+Field names must be atoms.
+
+When a native-record value is updated, its native-record
+definition is **not** consulted.
```
RaimoNiskanen: Same as a comment of mine above... Correct?

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+
+An update operation fails with a `{badrecord,Expr}` error if:
+
+* The definition of the record was not exported when the native-record
+  value was created, and it is now used outside its defining module.
+
+An update operation fails with a `{badfield,FN}` error if:
+
+* The native-record update expression references the field FN which is
+  not defined native-record value.
```
RaimoNiskanen: Suggested rephrase:
```
* The native-record update expression references a field FN which is
  not defined in the native-record value's captured definition.
```

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+fields:
+
+```erlang
+Expr#_{Field1=Expr1, ..., FieldN=ExprN}
+```
+
+Field names must be atoms.
+
+When a native-record value is updated, its native-record
+definition is **not** consulted.
```
RaimoNiskanen: Maybe more emphasis on what happens than on what not happens:
```
When a native-record value is updated, its captured
definition is consulted, **not** the current definition
in the defining module.

File: eeps/eep-0079.md [hunk 28]
```diff
@@ -0,0 +1,937 @@
...
+  value was created, and it is now used outside its defining module.
+
+* `Expr` does not evaluate to a native-record value of the expected
+  type `#Name` or `#Module:Name` (that is, it is either not a native
+  record at all or it is another native record).
+
+An anonymous update operation fails with a `{badfield,FN}` error if:
+
+* The native-record update expression references the field FN which is
+  not defined (in the structure definition).
```
RaimoNiskanen: `structure definition` -> `native-record definition`
?

I guess also an anonymous update consults the captured definition of whatever record `Expr` is, and suggest to state that here, for clarity, so:

`structure definition` -> `captured native-record definition`
?

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+
+```erlang
+#Name{Field1 = Expr1, ..., FieldN = ExprN}
+#Module:Name{Field1 = Expr1, ..., FieldN = ExprN}
+```
+
+Here, `Expr1`, .. `ExprN` are patterns, and Field names must be atoms.
+
+When a native-record value is matched, its native-record definition is
+**not** consulted.
```
RaimoNiskanen: Again, maybe say that `its captured native-record definition is consulted`?

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+```
+
+but guaranteed to always work and be more efficient.
+
+TODO: What should the name of the BIF be?
+
+We should also have a BIF that checks whether an instance of a native
+record referes to the current definition of the native record.
+
+TODO: What should the name of that BIF be?
```
RaimoNiskanen: `is_record_current/1`?

Such a BIF needs to consult the defining module.  What should it do if that module is not loaded?
Return `false` because obviously there is no record definition, currently?
Is it possible that a later module load will load the current definition, or is that impossible?
I guess we do not want the BIF to return `false` now and `true` later.  The other way around must be ok...

eeps/eep-0079.md
```diff
@@ -0,0 +1,937 @@
...
+Fetching the record index using the `#name.field` syntax is not
+supported, because there is no way it can actually be used, since neither
+ETS nor `element/2` will work with native records.
+
+### Native-record guard BIFs
+
+#### `is_record/3`
+
+The existing `is_record/3` BIF is overloaded to also accept a native record:
+
```
RaimoNiskanen, line 652: In general, maybe clarify which of these BIFs consult the captured definition, and which consult the defining module.

eeps/eep-0079.md
bjorng: We decided that in a previous OTB discussion. I'll add a note: "Note that the default values for tuple records can contain variables; we don't want that kind of complication here."

eeps/eep-0079.md
bjorng: I found one more.

eeps/eep-0079.md
bjorng: Yes, that is wrong. I will change to:

When a field from a native-record value is accessed, the **captured** native-record definition is consulted.

eeps/eep-0079.md
bjorng: I will rewrite to:

* The native-record value's captured definition states it as from a different module than the calling module, and it was *not exported* a the time of capture.

eeps/eep-0079.md
bjorng: Technically not wrong but ambiguous. I will change to:

When a native-record value is updated, the native-record definition captured when the record value was created is consulted.

eeps/eep-0079.md
bjorng: I will change to:

The native-record update expression references a field FN that is not defined in the native-record value's captured definition.

eeps/eep-0079.md
bjorng: I'll mark this part with NOT_REVIEWED_BY_OTB and copy your questions.

eeps/eep-0079.md
```diff
@@ -0,0 +1,798 @@
...
+* There is no corresponding native-record definition.
+
+* The native-record definition is not visible at the call site (it is
+  not exported).
+
+* The native record create expression references the field FN which is
+  not defined (in the structure definition).
+
+* No value is provided for a field FN and the native-record definition
+  has no default value for FN.
```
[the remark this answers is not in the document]
bjorng: Corrected.

eeps/eep-0079.md
```diff
@@ -0,0 +1,798 @@
...
+```erlang
+Expr#_.Field
+```
+
+This access operation fails with a `{badfield,Field}` error if:
+
+* The field `Field` is not defined in the native-record value.
+
+* The definition of the record was not exported when the native-record
+  value was created, and it is now used outside its defining module.
```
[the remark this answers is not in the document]
bjorng: Corrected.

eeps/eep-0079.md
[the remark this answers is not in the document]
bjorng: I've added a section named "Compile-time checking of records" that describes what the compiler checks.

eeps/eep-0079.md
```diff
@@ -0,0 +1,1012 @@
...
+```erlang
+#?MODULE:user{name = ~"John", city = ~"Stockholm"}
+```
+
+This is equivalent to the previous example that did not specify the
+module name.
+
+When creating a record value, the definition from the current version
+of the module will be used. There is no way to create a native record
+based of an old code generation.
```
garazdawi: Does this mean that this code:

```erlang
#user{name = ~"John", city = ~"Stockholm"}
```

when run in old code, would create a record using the new definition?

eeps/eep-0079.md
```diff
@@ -0,0 +1,1012 @@
...
+  loaded module on that node.
+
+#### Default values
+
+If no value is provided for a field, and there is a default field
+value in the native record definition, the default value is used. If
+no value is provided for a field and there is no default field value,
+a native record creation fails with a `{novalue,FieldName}` error.
+
+In OTP 29, the default value can only be an expression that can be
```
garazdawi, line 387: Is the goal to have arbitrary expression later on? IMO it should be a goal.

eeps/eep-0079.md
```diff
@@ -0,0 +1,1012 @@
...
+NOT_REVIEWED_BY_OTB.
+
+A native record creation is validated at runtime against the
+native record definition. Creation fails in the following
+cases.
+
+* If there is no corresponding native-record definition, creation
+  fails with a `{badrecord,Name}` exception.
+
+* If the native-record definition is not visible at the call site (it
```
garazdawi, line 441: I assume that this would include this example:

```erlang
-module(foo).
-record #bar{baz = 1}.
foo() -> ?MODULE:bar{}.
```

That is a creation of a non-exported record using the global syntax?

eeps/eep-0079.md
```diff
@@ -0,0 +1,1012 @@
...
+  is not exported), creation fails with a `{badrecord,Name}`
+  exception.
+
+* If the native record create expression references the field FN that
+  is not defined (in the native-record definition), creation fails
+  with a `{badfield,FN}` exception.
```
garazdawi: Should the record name also be part of this exception?

eeps/eep-0079.md
```diff
@@ -0,0 +1,1012 @@
...
+native-record value.
+
+When a field from a native-record value is accessed, the
+**captured** native-record definition is consulted.
+
+An access operation fails with a `{badrecord,Expr}` error if:
+
+* The native-record value's captured definition states it as
+  from a different module than the calling module, and it was
+  *not exported* a the time of capture.
```
garazdawi: Can you use `?MODULE:Name.Field` on a non-exported record in the same module?

eeps/eep-0079.md
```diff
@@ -0,0 +1,1012 @@
...
+
+```erlang
+-record #Name(TVar1, ..., TVarN) {
+             Field1 [= Value1] [:: Type1],
+             ...
+             FieldN [= ValueN] [:: TypeN]
+}.
+```
+
+`Name` and `FieldN` need to be atoms. `Name` is allowed to be used without quotes
```
garazdawi, line 136: Maybe list some examples of the things that you do need to quote, i.e. `_`, `1`, `=` etc?

eeps/eep-0079.md
bjorng: Yes.

garazdawi: Is that really a good idea? That would mean that a process running old code could create a record without a certain field and then crash when trying to match on that field. IMO local creation of native records should capture the definition of the running module and not the latest running module.

eeps/eep-0079.md
bjorng: We haven't decided yet. Supporting arbitrary expressions is tricky because record creating will have to execute expression in another module.

eeps/eep-0079.md
bjorng: In the current implementation, this creation succeeds. Currently, specifying the `?MODULE` is exactly equivalent to omitting the module name. Should we treat them differently, and if so why?

I will add your example to EEP so that we can discuss at the next OTB meeting.

eeps/eep-0079.md
bjorng: IMO, I think we should keep the exceptions simple but I'll add this as question to the EEP.

eeps/eep-0079.md
essen: > That would mean that a process running old code could create a record without a certain field and then crash when trying to match on that field

That would be pretty surprising behavior and make upgrades difficult. Old code shouldn't have to account for what the new code will look like. Only new code should handle what old code produces.

eeps/eep-0079.md
bjorng: Will do.

eeps/eep-0079.md
bjorng: Yes.

eeps/eep-0079.md
bjorng: > Is that really a good idea?

No, probably not.

> That would mean that a process running old code could create a record without a certain field and then crash when trying to match on that field.

I'm not sure that deleting a field that is used in the previous version of the code is a good idea anyway but I agree the currently implemented behavior can complicate code update.

> IMO local creation of native records should capture the definition of the running module and not the latest running module.

Yes, that seems to be safer. I think I could probably do that that change in time for RC2.

eeps/eep-0079.md
garazdawi: I've always imagined record creation to be similar to a function call in how it works. That is `#foo:bar{}` calls the `foo` module to create a `bar` record. Using this mechanism you can kind of also build "opaque" fields as an exported record could have defaults that are non-exported records.

```erlang
-export_record([bar]).
-record #bar{ private_data = #private{} }.
-record #private{ for_my_eyes_only = ~"hello" }.
```

eeps/eep-0079.md
garazdawi: > Should we treat them differently, and if so why?

Because it aligns with how function calls work, this making it more idiomatic Erlang IMO.

eeps/eep-0079.md
michalmuskala: Hmm... That's a good question. You can't do that with a non-exported function. Why should the behave differently?

eeps/eep-0079.md
bjorng: Yes, I will change the EEP to make native-records align with function calls.

eeps/eep-0079.md
garazdawi: One argument against allowing arbitrary expressions is that it would make it impossible to create a native record with such default values within a NIF.

bjorng:
I've pushed an updated version where I've adopted the view that native-records operations behave similar to calls (as suggested by @garazdawi).

* Operations that don't specify a module name (neither explicitly nor implicitly using `-import_record()`), references records in the current module. Exported or not doesn't matter.
* Operations that specify a module name (either explicitly or implicitly using `-import_record()` will only work if the record is exported (for creation), or was exported at the time the record value was created. That also applies to `?MODULE`; record operations (operating on the elements) with a module name will only work for exported records.

I've implemented most of these changes but will not make a PR until the EEP has been approved by the OTB.

eeps/eep-0079.md
```diff
@@ -0,0 +1,1024 @@
...
+
+For example, the record creation in the `foo/0` function will fail because
+the `#bar{}` record is not exported:
+
+```erlang
+-module(foo).
+-record #bar{baz = 1}.
+-export([foo/0]).
+foo() -> #?MODULE:bar{}.
+```
```
kikofernandez, line 308: `bar` record is not exported, but `bar` is defined in its defining module, 
https://github.com/erlang/eep/pull/81/changes#diff-6fbdcf9ebdcf73842e54c4a8c9e012724c2b3cb22b6816d07e1f65220501d078R139
and when the EEP says
> `foo/0` function will fail

is this a linter error, a runtime error if module `qux` calls `foo:foo/0` or an error if module `foo` calls internally function `foo/0`?

eeps/eep-0079.md
```diff
@@ -0,0 +1,1074 @@
...
+#users:user{id = 1, name = "Alice", city = "London"}
+```
+
+Printing of fields follows the field order. Whether a native-record
+value is exported is not visible through printing.
+
+### External term format
+
+External term format is extended to support serialization of
+native-record values.
```
michalmuskala: Should this section be expanded a bit on how the scheme will look like?
In particular given efficiency is a big part of the feature consideration, IMO compactness/efficiency of the encoding on-the-wire for distributed programming scenarios should be considered

eeps/eep-0079.md
bjorng: I will clarify that it will fail at runtime.

eeps/eep-0079.md
bjorng: I will add some details about the encoding.

bjorng:
I've pushed a version updated after the meeting of the OTP Technical Board. Most of the changes are clarifications of existing content, but OTB also decided that the `badfield` and `novalue` exceptions should also include the module namd and name of the record.

I've also added a brief description of the external format for native records.

The parts of the EEP on which OTB didn't reach a decision has been removed and saved in a branch (named in the EEP).

bjorng:
While implementing the things that had changed from the EEP, I realized that all of "Compile-time checking of records" is not implementable. Trying to use a tuple record lacking a definition has always been a compile error. If we don't have a record definition, we can't know whether a tuple record or native record was intended, so that must be an error.

I've pushed a fixup commit, which I hope will be the last change to this EEP. I hope to merge this tomorrow.

williamthome:
Hi! I noticed the Backward Compatibility section doesn't mention the impact on parse transforms and tools using abstract forms.

Native records will introduce new AST node types. Existing parse transforms that traverse the abstract forms may fail on code containing native record syntax (e.g., function clause errors on unknown nodes). Other EEPs that introduce new syntax typically address this — for example, EEP-0037 notes "Existing parse transforms might well fail on code containing the new form, but would work unchanged on code that does not" and EEP-0078 states "parse transforms or any tools using abstract forms, would need to be updated."

Should a similar note be added to this EEP's Backward Compatibility section?

bjorng:
Thanks, I've added a similar note.

eeps/eep-0079.md
```diff
@@ -0,0 +1,1082 @@
...
+
+The order of the field names as declared is preserved and part of the
+native record definition. When printed the fields are printed in the
+order defined.
+
+There is no defined maximum number of fields in a native record. Because of
+the need to define all fields in a record, it is very hard to accidentally
+create a record with one million elements.
+
+### `-import_record()`
```
garazdawi, line 168: One incompatibility of sorts that I've found is that `-record(....).` can be declared anywhere in the code, while `-import_record(...)` can only be declared before any function.

There exists code that for whatever reason includes `file.hrl` in in the middle of the code instead of at the top, so if you want to replace `file_info` with a native-record then those files break.

Moving the include up is of course a trivial fix, but depending on how many such you have it could be a lot of work.

eeps/eep-0079.md
```diff
@@ -0,0 +1,1082 @@
...
+
+As it can be seen from the next sections, working with native records outside of
+the defining module needs using fully qualified names of the native records:
+`#misc:user{}`, ... This quickly may become too cumbersome and verbose. The
+`-import_record` directive works similar to the `-import` and `-import_type` directives.
```
garazdawi: As you can do `-type foo() :: #foo{}`, should `-import_record` also import the record for types? IMO it should.

bjorng: For Dialyzer and other type checkers? Yes. I don't see why they wouldn't do that.

eeps/eep-0079.md
bjorng: Yes, it should be allowed to be anywhere in the file. I've added a paragraph about that, and will post a message to the OTP team to see whether there any objections.

eeps/eep-0079.md
garazdawi: I was apparently testing with a non-master compiler. It already does what I was after. i.e. that this works:

```erlang
-module(test).

-import_record(foo, [foo]).

-type foo() :: #foo{}.
```

sorry for the noise.

PR #82, opened 2025-12-09
Add EEP for partially applied functions

josevalim:


RaimoNiskanen:
How would `fun foo({ok,_})` work?  Can an argument be any pattern?

Can an argument be an unbound variable?  `fun foo(X)`

Can the fun name be a bound variable?
```erlang
    F = foo,
    fun F(_),
    F(bar),
```
or does it have to be a literal function name `foo` or `mod:foo`?

This would be allowed, right?
```erlang
foo(Y) -> Y-1.
bar(X) ->
    F1 = fun Foo(X) -> X+1 end, % Arity 1
    F2 = fun foo(X),            % Arity 0
    F3 = fun foo/1,             % Arity 1
    {F1(X), F2(), F3(X)}.
```
`bar(3) -> {4,2,2}`
where on line F1 there would be warnings that X is shadowed and that Foo it not used.

Maybe what I am getting at is that it is a bit subtle that F2 defines a function body (with a hidden header) where F1 and F3 defines a function header (and for F1 also a body).

josevalim:
Thank you @albsch and @RaimoNiskanen!

> How would fun foo({ok,_}) work? Can an argument be any pattern?

I would say `{ok, _}` would not work. An argument is either a literal or a variable.

> Can the fun name be a bound variable?

Since `fun F/1` is not valid today, I assume `fun F(Arg1, ...)` shouldn't be valid either. But given `fun Mod:Fun/Arity` is valid, I assume `fun Mod:Fun(Arg1, ...)` should be implementable at runtime with the properties I outlined here, as long as Mod and Fun are bound variables. I will clarify it.

> Maybe what I am getting at is that it is a bit subtle that F2 defines a function body (with a hidden header) where F1 and F3 defines a function header (and for F1 also a body).

I agree they feel a bit too close. There are some trade-offs that could be made here:

1. Don't allow `fun some_mod:some_fun(Args)` or `fun some_fun(Args)` when there are no placeholders. After all, if there are no placeholders, it means they could be represented as a `fun some_mod:another_fun/0` where `another_fun` calls the original `some_fun` with all arguments statically. FWIW, Elixir has this restriction.

2. Only allow remote partially applied functions, so `fun foo(_, ok)` doesn't work, only `fun some_mod:foo(_, ok)`, but I am afraid it will lead to developers doing external calls when a local call would suffice for the syntax convenience.

3. Do nothing, since named anonymous functions are not common anyway.

Any thoughts?

josevalim:
> F = fun foo(X, _) would be a local fun with X in its environment, this exists already, right?

Yes, I think this one wouldn't need runtime changes, since local functions are not external/serializable/persistent (we need a better way to describe those...).

> But F = fun ?MODULE:foo(X, _) would be a new thing - an anonymous export entry with an arity and an environment.(?) 

Unfortunately I cannot speak about the implementation details. I assume one option is to extend the existing external function types to have a field that points to its partially applied arguments? For all existing `fun Mod:Fun/Arity`, this field is empty.

jhogberg:
> Or should it be a new term type that refers to an export entry?

As per the current fun header definition:

```c
/* Fun objects.
 *
 * These have a special tag scheme to make the representation as compact as
 * possible. For normal headers, we have:
 *
 *     aaaaaaaaaaaaaaaa aaaaaaaaaatttt00       arity:26, tag:4
 *
 * Since the arity and number of free variables are both limited to 255, we can
 * fit them both into the header word.
 *
 *     0000000keeeeeeee aaaaaaaa00010100       kind:1,environment:8,arity:8
 *
 * Note that the lowest byte contains only the function subtag, and the next
 * byte after that contains only the arity. This lets us combine the type
 * and/or arity check into a single comparison without masking, by using 8- or
 * 16-bit operations on the header word. */

#define FUN_HEADER_ARITY_OFFS (_HEADER_ARITY_OFFS + 2)
#define FUN_HEADER_ENV_SIZE_OFFS (FUN_HEADER_ARITY_OFFS + 8)
#define FUN_HEADER_KIND_OFFS (FUN_HEADER_ENV_SIZE_OFFS + 8)

#define MAKE_FUN_HEADER(Arity, NumFree, External)                             \
    (ASSERT((!(External)) || ((NumFree) == 0)),                               \
     (_TAG_HEADER_FUN |                                                       \
     (((Arity)) << FUN_HEADER_ARITY_OFFS) |                                   \
     (((NumFree)) << FUN_HEADER_ENV_SIZE_OFFS) |                              \
     ((!!(External)) << FUN_HEADER_KIND_OFFS)))
```

There's a few places that assume that external funs don't have an environment at the moment, notably in the external term format, but otherwise it would be easy to let them have an environment too.

I think it's possible to make these terms round-trip to old nodes and back without loss, by using `NEW_FUN_EXT` with an impossible signature (`old_uniq` is always derived from `uniq` IIRC). They wouldn't be callable on the old nodes, appearing as an anonymous function that always `badfun`'s when called, but it would work when passed onward to a new node.

eeps/eep-00XX.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+The `_` variable denotes placeholders, which are arguments
+that have not yet been provided (the use of `_` is a proposal,
+the exact notation can be changed). The number of placeholders
+dictate the arity of the function and they are provided in order.
+For example:
+
+```erlang
+fun hello(_, world, _)
```
frazze-jobb: would partial application in steps be interesting or just confusing?
Fun1 = fun my_foo( _, b, _).
Fun2 = fun Fun1(_, c),
Res = Fun2(a),  %% fun my_foo(a,b,c)

josevalim: @frazze-jobb given `fun Foo/2` is not valid today, I believe we could allow such without introducing any ambiguity, but it does add to the concerns described in the "visual cluttering" section.

michalmuskala: Actually I believe that right now the EEP does not address the case of `fun Foo(_, c)` where `Foo` is a variable. It does say that this doesn't work fo call a local function if `Foo` is an atom, but it doesn't specify if this should fail to be an function application if `Foo` is bound to another fun.
If this indeed is valid and `Foo` is bound to a "serialisable" fun, should an expression like this produce a "serialisable" fun as well?

josevalim: Yeah, that's my read of the situation too. I will add it later to the EEP as a possible extension.

eeps/eep-00XX.md
```diff
@@ -0,0 +1,317 @@
...
+{Mod, Fun, Args} = MFArgs
+apply(Mod, Fun, [SomeValue | Args]).
+```
+
+One of the main reasons MFArgs exist is because anonymous functions
+which close over an existing environment cannot be serialized across
+nodes nor be persisted to disk, so when dealing with distribution,
+disk persistence, or hot code upgrades, you must carefully stick with
+MFArgs. Similarly, configuration files do not support anonymous
+functions, and MFArgs are the main option.
```
rickard-green: Passing an anonymous function with an environment over the distribution works. I must be misunderstanding something?

josevalim: Poor phrasing on my end, I should clarify it only works if you have the exact same module version (which, unless you declare a version, will change even with minor changes to the source). I will improve that on a future pass to the EEP.

jhogberg:
> @jhogberg one thing of note is that this new environment maps arguments to positions. They are still ordered, but they can have gaps, such as `fun foo:bar(_, arg1, _, arg2)`. I am unsure if that complicates the format in any way...

Not by much, the implementations I had in mind are relatively easy to implement with reasonable performance.

eeps/eep-00XX.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+The placeholder must always appear in the position of an argument,
+it cannot be nested inside construct. The following is not allowed:
+
+```erlang
+fun hello({_, world}, _)
+```
+
+Furthermore, all arguments that are not placeholders are evaluted
```
Maria-12648430:
```suggestion
Furthermore, all arguments that are not placeholders are evaluated
```

eeps/eep-00XX.md
```diff
@@ -0,0 +1,318 @@
...
+they are provided in order. For example, this is a two arity
+function:
+
+```erlang
+%% two arity function
+fun hello(_, world, _)
+```
+
+The placeholder must always appear in the position of an argument,
+it cannot be nested inside construct. The following is not allowed:
```
Maria-12648430:
```suggestion
it cannot be nested inside a construct. The following is not allowed:
```

Maria-12648430 (changes requested):
Some typos, some questions/suggestions/opinions :)

Maria-12648430 (approved):

eeps/eep-00XX.md
```diff
@@ -0,0 +1,318 @@
...
+
+In particular, `erlando` does not require the `fun` prefix, so one can
+write:
+
+```erlang
+maps:get(username, _)
+```
+
+The lack of a prefix makes it harder to spot when a function is created
+and also leads to visual ambiguity, such as in the code below:
```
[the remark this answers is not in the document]
josevalim: @Maria-12648430 it has been a long discussion, so I believe your comment means you prefer the `fun` prefix for clarity?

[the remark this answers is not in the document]
Maria-12648430: Yes, absolutely.

Without it, it would be harder to distinguish the creation of a partially applied function from a call to a function when skimming through code.

When I see `fun`, I know that I get a function. Without it, I will have to read more closely if there is a `_` in the arguments or if there is a `/X` at the end.
_(And now that I think about it, `/X` could also mean "divide the result of a call by X". This ambiguity would not only make it harder to distinguish calls and partial applications, it would also mean that a full partial application is not even _possible_ since it is impossible to tell whether sth like `add(1, 2)/0` is a full partial application of `add/2` or the result of a call to `add/2` divided by zero.)_

Without the `fun`, basically anything that _looks_ like a call could also be a partial application. You have to examine it, _closely_, to be sure.

Without the `fun` in front, like this...
```erlang
Prod = lists:foldl(fun(E, Acc) ->
                       Acc * E
                   end,
                   1,
                   _)/1.
```
... it is not clear at a glance that what ends up in `Prod` here is a function of arity 1, instead of some result of a `lists:foldl` call. You have to spot the `_` buried in the arguments list to be sure, since even the `/1` at the very end is ambiguous and could mean "divide by 1".

Now, with `fun` in front, it is really obvious that you get a function there:
```erlang
Prod = fun lists:foldl(fun(E, Acc) ->
                           Acc * E
                       end,
                       1,
                       _)/1.
```

Those are simple examples. Imagine this with the fold function a partially applied function, or something more complex, the sky is the limit ;)

In short: all those problems do not exist when the `fun` prefix is used 🤷‍♀️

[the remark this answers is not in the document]
josevalim: Thank you! And do you have any preference on having an explicit arity vs have it inferred by counting the number of `_`?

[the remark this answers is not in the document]
Maria-12648430: Hm, my preference would be to have explicit arity, but this is just a personal peference, nothing grounded in perceived problems or benefits.

Maria-12648430:
Btw, unless I have missed something, partial application of anonymous (vs named) functions has not been discussed?

Would sth like this be possible?

```erlang
Add = fun(A, B, C) -> A + B + C end.
AddOne = fun Add(_, _, 1)/2.
AddOneAndTwo = fun AddOne(_, 2)/1.
```

josevalim:
@Maria-12648430 from my point of view that should just work, as we should allow any function application inside `fun`.

eeps/eep-00XX.md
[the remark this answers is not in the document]
essen: I also highly prefer the `fun` style, especially because otherwise it's easy to confuse with catch-all matching. The `fun` makes it more obvious.

On the other hand I don't think the arity helps much once you have the `fun` and I can easily see it become annoying when refactoring and adding or removing arguments.

